PR #29, opened 2020-03-26
add version function, and update build instructions

Milerius:


ca333:
i changed the base to `dev` @Milerius

Issue #25, opened 2020-03-22
Mouse hover is not captured when cursor is on the graph


naezith:
![image](https://user-images.githubusercontent.com/6732486/77258267-a05da600-6c8a-11ea-841e-0f35f5ec401a.png)

Hover is ignored when cursor is on graph.

ChartView component is used https://doc.qt.io/qt-5/qml-qtcharts-chartview.html

csharpee:
The only fix I am able to come up with at the moment is a temporary fix. Keep in mind, I am very new to Qt5.

I set the 7d Trend to layer.enabled: 1 and z: -1, then set the row opacity: .75.
It works, although the chart isn't as vivid or bright because it is peaking behind the opacity level set on the rows. 

The only other option could be a full layout change to have more control over each column.. 
I have sent a tweet to qtcreator; maybe they will have another option for us.

![Fix_HoverBug_Chart_Qt5-2020-03-28_19 49 58](https://user-images.githubusercontent.com/18726788/77836562-a8f02980-712d-11ea-882a-a710b8a8657d.gif)

tonymorony:
At the moment it works fine on OSX/Windows 0.1.4 so I believe this one might be closed

PR #41, opened 2020-03-29
Gui fixes

csharpee:
- There is a display bug on screen height < 800 with orderModel.qml. The model cuts off the bottom to use the close or view at explorer. 
- Used golden ratios for sidebar right and coinlist bar.
- Updated a fix for the hover over chart related to issue: #25

esy92-6d7nu


csharpee:
Thanks for the feedback, I have updated the changes back to original look. 
It is spacey; the reason for the row width changes was because the chart is overlapping each other and caused other problems with the z index hack; thebackgroundColor: "transparent" displays this when disabled. 

The space for the sidebars I personally liked, although the space for the rows did feel a bit bloated and the KMD log and navigation should be centered.

csharpee:
I am closing this pull request and will update the values in another iteration.

PR #61, opened 2020-04-01
added ILN + corresponding icon

BloodyNora:


ca333:
according to our info ILN tagged `dead` and getting removed from dPoW soonish

PR #80, opened 2020-04-04
cleanup coins.json

cipig:


cipig:
Yes, it removes all data that is already present in https://github.com/KomodoPlatform/coins/blob/master/coins, because atomicDEX-Pro does actually use the data from there, not from this config.

ca333:
> because atomicDEX-Pro does actually use the data from there, not from this config.

just for now and only for the spv mode - we haven't enabled the `native` mode yet. We will certainly use one config file towards `beta`. I believe changing the current `coins` structure to serve as a "central" config point makes most sense - we will sort this out soonish. Will keep you updated.

PR #105, opened 2020-04-16
feat(mnemonic): add mnemonic validation

Milerius:


naezith:
It wasn't working. It was triggering the `assert`. I removed the entropy part and just split the string and feed the string vector into validate function. It works but might wanna polish the split part.

PR #106, opened 2020-04-19
feat(dependencies): better dependencies list

Milerius:


dchmelik:
For GNU/Linux (appimage is for it in general, not only Ubuntu) there should be a dependency list.

PR #112, opened 2020-04-25
Add SUPERNET

zatJUM:


cipig:
```
    "rpcport": 11341,
    "mm2": 1,
```
is not needed, these are taken from the coins file

zatJUM:
Updated

PR #126, opened 2020-04-30
feat(ci_tools): better size for dmg package

Milerius:


ci_tools_atomic_dex/dmg-packager/package.sh
```diff
@@ -62,7 +62,7 @@ popd
 # figure out how big our DMG needs to be
 #  assumes our contents are at least 1M!
 SIZE=`du -sh "${STAGING_DIR}" | sed 's/\([0-9\.]*\)M\(.*\)/\1/'` 
-SIZE=`echo "${SIZE} + 1.0" | bc | awk '{print int($1+0.5)}'`
+SIZE=`echo "${SIZE} + 10.0" | bc | awk '{print int($1+0.5)}'`
```
ca333, line 65: +10M could be overstated - something between 2-5 should be sufficient

PR #136, opened 2020-05-04
0.1.3

ca333:


ca333:
> 2. negative BTC input when balance is too low (need an additional check in the frontend to be sure to not call backend function if we don't have balance > minimal_trade_balance)

likely makes sense to add a sanity check to ensure this can not happen backend-side @Milerius

naezith:
> 1. Can't paste ETH address when sending

https://github.com/KomodoPlatform/atomicDEX-Pro/pull/157

naezith:
> 2. negative BTC input when balance is too low (need an additional check in the frontend to be sure to not call backend function if we don't have balance > minimal_trade_balance)

https://github.com/KomodoPlatform/atomicDEX-Pro/pull/159

Milerius:
@ca333 Yeah that's added on both, let's merge #160 #159 #157 and we are ready to bundle/release

ca333:
@tonymorony execute full test and provide us QA signoff

PR #147, opened 2020-05-13
feat(average_swap_time): start swap average feature

Milerius:


Milerius:
No merging, it's for 0.1.4

PR #149, opened 2020-05-13
Recover funds button for Swap History

naezith:
Requires test, I don't have any recoverable ones. @tonymorony @Milerius


Milerius:
good for me too !

PR #174, opened 2020-05-21
Change syncing text

naezith:
`Syncing %n TXs...` to `Syncing... Current block: %n`


tonymorony:
I think just "syncing" is not informative, technically it's:
"Scanning blocks for transactions", or something like a "Syncing tx history"

I propose to deprecate this element in the new UX and left little bulb somewhere in the screen corner which is green when whole blockhain is scaneed for txs and orange when its not. Because now ino this info more confusing user than providing useful information

PR #180, opened 2020-05-21
More reliable price calculation

naezith:
`rel_amount / base_amount` price calculation is done in backend with boost.

This branch includes `trade_amount_function` branch.


atomic_qt_design/qml/Constants/API.qml
```diff
@@ -43,6 +43,8 @@ QtObject {
...
 
+        get_price_amount: (base_amount, rel_amount) => (parseFloat(rel) / parseFloat(base)).toFixed(8).toString(),
```
ca333, line 46: need to use decimal for arithmetic operations such as divisions

naezith: This is Mock API, used only in Design Studio.

naezith:
Preffered price is now 1:1 same as Orderbook price string. for `2 / 3` case it's `0.666666666666666666666666666666666666666666666666666666666666666666666666667`.

That solves the problem when you select a price from orderbook and it wasn't matching.

And user-written 3 to 2 is calculated with boost, maximum of 50 digits.  `0.666...7`< again.

However if user enters 100.5 to base input and 15 to rel input. Then submit that trade, mm2 returns `14.999999999999999999999999` kind of a value for rel side.

naezith:
For this trade 
image

Taker order mm2 returns shows 
image

When it becomes maker order, mm2 sends this
image

I limited 8 digits while showing it, and it looks like this now: 
image

And when it is maker, since that 1 is within the 8 digits, it's visible.

image

Even with price calculation with 50 digit precision, this happens. 

However, we tried this trade, 77.777 to 100.5 from both sides, and orders matched and trade completed successfully.

image

PR #187, opened 2020-05-26
improvements(fees): adapt get_trade_fee logic

Milerius:
bug reported by cipi:
`this is making problems on the parity nodes (all erc20 tokens share the same servers)`


artemii235:
Could you please clarify for what purpose `get_trade_fee` method is called every 30 seconds?

Milerius:
@artemii235 Everything in the backend is using cache parallels map.

Main thread just pick values from those parallels map. (balance, fees, transactions)

tonymorony:
I think we should call this when it's needed: on trade event
 and for needed coin: coin used in trade

now it looks like a spam because calling that for all activated coins permanently and on any page

Milerius:
It's already the case @tonymorony the orderbook is refreshed only when you are on the page of the trade event

tonymorony:
I've seen that orderbook polling on trading page only. We're discussing `get_trade_fee` I believe - I never mentioned orderbook.

Milerius:
need an upgrade from @naezith 

(set_current_orderbook take now a rel ticker (its for get_trade_fee))

instead of

Capture d’écran 2020-05-27 à 07 13 59

please write `Calculating trading fee` instead of `NaN` and poll get_trade_infos (this is not processing any rpc call) every seconds until you get a value.

artemii235:
> So if i get you guys there is only few rpc call that cause us trouble: get_trade_fee, tx_history, orderbook.

Actually only `get_trade_fee` because MM2 makes request to coins RPC servers when running it. E.g. `blockchain.estimatefee` in Electrum mode or request to ETH gas station API for ETH/ERC20. `orderbook` and `my_tx_history` do not trigger requests to coin RPCs so these may be called as frequent as required.

PR #240, opened 2020-06-11
add NavCoin (NAV)

cipig:
image


Milerius:
broken by  #243

cipig:
just merge this one first, than #243 and it should be fine
but no problem, i can simply remake this

PR #248, opened 2020-06-12
Fix swap details taker/payment ID namings and 0x prefix to ERC-20 URL

naezith:
Fixing: https://github.com/KomodoPlatform/atomicDEX-Pro/issues/246

    "Maker Payment ID" is actually the txid of "Taker Payment"
    "Taker Payment ID" is actually the txid of "Maker Payment Spent"
    the link to erc20 blockexplorer does not work because txids on etherscan start with 0x

@cipig Can you please test this?


cipig:
everything fixed, maker order, taker order and erc20 blockexplorer link
thanks

PR #275, opened 2020-06-21
fix(balance): remove usage of locked funds

Milerius:


Milerius:
Reopenning after the meeting, @artemii235 Confirmed that the field will be removed !

PR #284, opened 2020-06-24
improvements(security): using a configuration file instead of passing…

Milerius:
… through the command line

### DIFF
--- a/src/atomic.dex.mm2.cpp
+++ b/src/atomic.dex.mm2.cpp
@@ -604,10 +604,16 @@ namespace atomic_dex
         const auto tools_path = ag::core::assets_real_path() / "tools/mm2/";
 
         nlohmann::to_json(json_cfg, cfg);
+        fs::path mm2_cfg_path = (fs::temp_directory_path() / "MM2.json");
 
-        const std::array<std::string, 2> args = {(tools_path / "mm2").string(), json_cfg.dump()};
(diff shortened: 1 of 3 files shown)


src/atomic.dex.mm2.cpp
```diff
@@ -617,12 +623,12 @@ namespace atomic_dex
...
         }
 
-        m_mm2_init_thread = std::thread([this]() {
+        m_mm2_init_thread = std::thread([this, mm2_cfg_path]() {
             using namespace std::chrono_literals;
             // loguru::set_thread_name("mm2 init thread");
 
             const auto wait_ec = m_mm2_instance.wait(2s).second;
-
+            fs::remove(mm2_cfg_path);
```
Milerius, line 631: Here @artemii235 I highly recommend from MM2 core to delete the file themselves once you have parsed it, It's will be much more faster than me, i have no idea when you don't need the file anymore

Milerius: We can pass through another env variable if you prefer:

`MM_TEMP_CONF_FILE` (if you want to keep the other behavior)

Milerius:
I suggest that we merge this pull request as is. Improvement possibilities are possible via the core, I discussed it with @artemii235 , and the solution of this pull request is anyway more secure than the previous one.

PR #301, opened 2020-06-30
Trade page negative amount message

naezith:
image


naezith:
I just did a change, MAX button and the input field are being disabled if input is negative.

PR #304, opened 2020-07-01
Handle when sell volume is not enough to pay fees

naezith:
I'd have couple requests:
- `fees_needed` can be named to `amount_needed`, it's the one displayed in the warning
- Hardcoded `0.00777` value is now used in backend, since I use it in front-end too, it would be more less error-prone to have it hardcoded in single place, backend.

image


ca333:
> I'd have couple requests:
> 
> * `fees_needed` can be named to `amount_needed`, it's the one displayed in the warning
> * Hardcoded `0.00777` value is now used in backend, since I use it in front-end too, it would be more less error-prone to have it hardcoded in single place, backend.


yes - this needs handling in backend. @Milerius for the next iteration take over the sanity check calculations from the mobile app logic - same applies for total fees, et cetera.

PR #338, opened 2020-07-06
update BCH

cipig:


Milerius:
blocked by #336 becase we update config file to 0.2.0

PR #352, opened 2020-07-11
Privacy UI

naezith:
Implements https://github.com/KomodoPlatform/atomicDEX-Pro/issues/332

Half of this feature was implemented at `naezith_dev` already. It is complete with these changes.

The DEX page does not have privacy.

I can hide more stuff if requested, here is the current look:

image
image
image
image
image

### DIFF
--- a/atomic_qt_design/qml/Components/DefaultText.qml
+++ b/atomic_qt_design/qml/Components/DefaultText.qml
@@ -11,7 +11,7 @@ Text {
     font.family: Style.font
     font.pixelSize: Style.textSize
     color: Style.colorText
-    text: General.privacy_mode ? "*****" : text_value
+    text: privacy && General.privacy_mode ? "*****" : text_value
     wrapMode: Text.WordWrap
 }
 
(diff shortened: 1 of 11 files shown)


tonymorony:
The balances are visible atm trough base coin selector:
image

I think it's logical for this feature to hide these numbers as well

naezith:
Amount input and MAX slider are still visible. DEX section won't be usable if we do this. Binance does not have an option to hide stuff at the trade page.

These are the reasons why I didn't take that action.

tonymorony:
> Amount input and MAX slider are still visible. DEX section won't be usable if we do this. Binance does not have an option to hide stuff at the trade page.
> 
> These are the reasons why I didn't take that action.

I see, in such case there is no much sense to hide information on Orders and History pages

I think it might be worth to consult with @ca333 - I realised that don't understand on 100% that particular feature practical sense

naezith:
> I see, in such case there is no much sense to hide information on Orders and History pages

I think that's still good because you can still see the date, result, and names/icons of the coins.

Milerius:
I'm merging this one, at the worst case we will open enhancement issues. (The main feature is present and working, and the following change are needed to implement addressbook)

ca333:
> I'm merging this one, at the worst case we will open enhancement issues. (The main feature is present and working, and the following change are needed to implement addressbook)

it seems this was merged although a question from https://github.com/KomodoPlatform/atomicDEX-Pro/pull/352#issuecomment-657203489 was still open - plz ensure this does not happen again. Open questions or outstanding approvals from Tony are equal and essential requirement for merging @Milerius !

PR #354, opened 2020-07-12
Fix install readme for OSX

dathbezumniy:
```
-README.md upd:
 +libbitcoin installation instructions
 +libbitcoin-system installation instructions
```

### DIFF
--- a/src/atomic.dex.app.tests.cpp
+++ b/src/atomic.dex.app.tests.cpp
@@ -1,0 +1,28 @@
+/******************************************************************************
+ * Copyright © 2013-2019 The Komodo Platform Developers.                      *
+ *                                                                            *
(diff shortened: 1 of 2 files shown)


dathbezumniy:
a381bc7, 2de58a2 please exclude these commits from merging, not sure how to do that at this point...

src/atomic.dex.app.tests.cpp
```diff
@@ -0,0 +1,30 @@
+/******************************************************************************
```
Milerius, line 1: This files should not be in this pull request

Milerius:
You probably want to sync your fork first, and do a pull request from dev branch of this fork

src/atomic.dex.app.tests.cpp
dathbezumniy: yea, I might have created some files during my testing session and forgot that they will be committed to this pull-request.

PR #357, opened 2020-07-13
feat(addressbook): implement addressbook feature

Milerius:


tonymorony:
image

contact name length is unlimited atm - I propose to set its limit to 50 chars

PR #361, opened 2020-07-19
feat(update_notification): finalize logic

Milerius:


tonymorony:
found one edge case it seems

```
  "new_version": "0.2.0",
  "version_num": "020",
  "status": "required",
```

I've set json to that value so @naezith will be able to check. Otherwise seems good

naezith:
What is the problem exactly?  Slyris handles it and gives me "update_needed" field. @Milerius

tonymorony:
> What is the problem exactly? Slyris handles it and gives me "update_needed" field. @Milerius

problem is that version is 0.2.0 and in json 0.2.0 as well but I got update screen which I can't skip

tonymorony:
in case of  ` "status": "required"` flag user should get that unskipable screen only when user's app version is < than  `version_num`

PR #365, opened 2020-07-20
Prioritize KMD/BTC coins at DEX

naezith:
With our current DEX ticker `ComboBox` model is automatically choosing the first one at the list. 

This change detects the model reset and checks if `KMD` is available, chooses that, if not, chooses `BTC`.

Normally, changing `Sell/Base`coin resets the `Receive/Rel` coin, and the alphabetically first coin in the list is being selected. This is still same, but now `Rel` side becomes `KMD` in this situation. Unless `Base` side is `KMD`, in that case, `Rel` side becomes `BTC`. 

Please check for QML errors which would look like ` qrc:/atomic_qt_design/qml/Constants/API.qml:0: RangeError: Maximum call stack size exceeded.`


tonymorony:
This version do not draw candlestick charts. Don't see anything relevant in logs.

PR #367, opened 2020-07-22
feat(orders/history): start switching to model view of history/orders

Milerius:


atomic_qt_design/qml/Exchange/Exchange.qml
```diff
@@ -161,67 +166,40 @@ Item {
...
-        return status === status_swap_matching ? qsTr("Order Matching") :
-                status === status_swap_matched ? qsTr("Order Matched") :
-                status === status_swap_ongoing ? qsTr("Swap Ongoing") :
-                status === status_swap_successful ? qsTr("Swap Successful") :
-                                                        qsTr("Swap Failed")
+    function getStatusColor(status) {
+        return status === "matching" ? Style.colorYellow :
+               status === "matched" ? Style.colorOrange :
+               status === "ongoing" ? Style.colorOrange :
+               status === "refunding" ? Style.colorOrange :
```
ca333, line 180: why not "group" these with an OR?

atomic_qt_design/qml/Exchange/History/History.qml
```diff
@@ -63,18 +41,12 @@ Item {
 
         SwapList {
             title: API.get().empty_string + (qsTr("Recent Swaps"))
-            items: getRecentSwaps()
+            items: API.get().orders_mdl
```
Milerius, line 44: why this is not using the proxy ?

atomic_qt_design/qml/Exchange/Exchange.qml
naezith: It's cleaner to see and modify which status has what. It's not a performance sensitive code part so I prefer clarity. This way they are same as other functions like `getStatusText` and `getStatusStep`. Those won't have that `OR` for those states. I can still do that if you want, using a switch is also possible but this code will be shorter.

atomic_qt_design/qml/Exchange/OrderModal.qml
```diff
@@ -43,11 +47,9 @@ DefaultModal {
...
             Layout.topMargin: 20
             font.pixelSize: Style.textSize3
-            visible: getStatus(details) !== status_swap_not_swap &&  // Is order
-                     (details.events !== undefined || // Has events, ongoing or
-                    details.am_i_maker === false) // Taker order with no events
-            color: visible ? getStatusColor(details) : ''
-            text_value: API.get().empty_string + (visible ? getStatusTextWithPrefix(details) : '')
+            visible: !details ? false : details.is_swap || !details.is_maker
+            color: !details ? "white" : visible ? getStatusColor(details.order_status) : ''
+            text_value: API.get().empty_string + (!details ? "" : visible ? getStatusTextWithPrefix(details.order_status) : '')
```
Milerius, line 52: This part is hard to read do to an excessive usage of the `ternary operator`, is it possible to make it more readeable ?

Milerius:
Overall comment:

In general, the use of the ternary operator is interesting in certain cases, and I understand that it drastically reduces the quantity of code but to the detriment of reading and understanding the code. In this sense I prefer that beyond two ternary operators chained on the same line you use:

a switch
or normal conditions

tonymorony:
view at explorer button producing tx not found for both maker/taker transactions

image

image

I've caught tx id it's trying to open and the case is that maker and taker payment mixed places (RICK txid app trying to open in MORTY explorer and vice versa)

atomic_qt_design/qml/Constants/General.qml
```diff
@@ -192,62 +192,6 @@ QtObject {
         return o === undefined ? def : o[field]
     }
```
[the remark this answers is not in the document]
naezith: Removed getField function

tonymorony:
quite tiny issue, when you start swap counter switching from 0 to 1 but when swap finishes its 1 still until you switch tab back and forth 

image

atomic_qt_design/qml/Exchange/Exchange.qml
naezith: Turned them to switch cases.

atomic_qt_design/qml/Exchange/History/History.qml
naezith: It's not the `ListView`, it's the wrapper. This takes the `orders_mdl` so it can use `.length` inside, and feeding `proxy` into the real `ListView`.

atomic_qt_design/qml/Exchange/OrderModal.qml
naezith: I see. I'm trying to code keep property bindings not having functions if possible. `!details` part is just checking if the object is healthy. The rest of it is the real code. I can move them to the new line, that would help a lot.

naezith: Moved real logic to the next line at property bindings.

naezith:
https://github.com/KomodoPlatform/atomicDEX-Pro/issues/323 is fixed here

PR #395, opened 2020-07-28
DEX Privacy and Linux BTC Symbol

naezith:
- Ubuntu is not able to show ₿ so Linux build uses ฿ instead.
- Privacy filter for DEX Sell coins list

image


tonymorony:
yay 👍 
image

PR #402, opened 2020-07-31
feat(multiple_currencies): multiple currencies feature

Milerius:


tonymorony:
It would be great to refresh displaying values on price selection in settings.

image
image

to reproduce:

1) Wallet (track price value)
2) Settings -> change currency
3) Wallet 

Ticker will change but actual price not until you switch to another coin

PR #403, opened 2020-08-01
High dpi screen

Milerius:


tonymorony:
13" 1920x1080 175% zoom:

image

PR #412, opened 2020-08-04
feat(dpow_conf): add backend for dpow confs

Milerius:


tonymorony:
Current todos:

1) please ensure that this logic is imlemented:
https://github.com/ca333/komodoDEX/issues/690#issuecomment-668496869
https://github.com/ca333/komodoDEX/issues/690#issuecomment-669087491

( dPoW can be enabled only for dPoW protected chains `incoming` transaction )

2) dPoW on/off is boolean thing, so only upper switch is needed, check mark selector is redundant

image

3) Change labels  "Enable dPoW Configurations" "Enable Normal Configurations" to "Enable Komodo dPoW security" "Change required confirmations"

4) Add (i) icon and on click show popup with text - this text will be explanation of this protection, required warning and etc. Fill text by placeholder for now, then we'll provide text approved by marketing dept

image

5) Also there should be confirmation warning if user is going to start swap with disabled dPoW protection if it's potentially possible to enable it (Like a "Are you sure to start the swap with dPoW protection" ? )

PR #440, opened 2020-08-20
feat(pin_cfg): add random for pin_cfg and apply factor to get_balance

Milerius:


Milerius:
erc/eth history rdy

tonymorony:
> erc/eth history rdy

erc20/eth tx history from endpoint works fine on my side. thanks!

tonymorony (approved):

PR #457, opened 2020-09-02
feat(price_aggregator): add global price aggregator and band oracle

Milerius:


smk762 (approved):
Encountered issue with restart failing after setting non-USD fiat .
Appears to be resolved in https://github.com/KomodoPlatform/atomicDEX-Pro/pull/458 which is a continuation of this PR, so will approve with an asterix.

PR #458, opened 2020-09-03
New wallet model

Milerius:


smk762 (approved):
Tested 
- tx history scrolling
- notifications
- fiat update on wallet page
- clear history on disconnect
- smoke test (swaps, aur claim etc)
- ETH history endpoint
- band oracle links

Much more responsive!

PR #484, opened 2020-10-03
Update atomicdexpro_ci.yml

smk762:
get nim without scoop


smk762:
Confirmed working in https://github.com/KomodoPlatform/atomicDEX-Pro/runs/1203181399?check_suite_focus=true
Currently uses direct download from https://nim-lang.org/download/nim-1.2.6_x64.zip but can be changed to locally hosted givena URL.

PR #487, opened 2020-10-09
feat(rename): rename to atomicdex-desktop

Milerius:


Milerius:
@naezith please do the require change in front-end to match the renaming to atomicdex-desktop

smk762 (dismissed):
Multi OS tests completed without issue.

PR #503, opened 2020-10-16
feat(band_oracle): update to last band oracle protocol

Milerius:


smk762 (approved):
icons sighted, link verified and fiat/kmd/btc values confirmed.

Issue #504, opened 2020-10-16
[FEATURE REQUEST]: Display fiat equivalent of amount and fees when sending coins


SylEze:
**Is your feature request related to a problem? Please describe.**
The fiat equivalent of the amount and the fees are not displayed when sending coins.

**Describe the solution you'd like**
First we need to change the request's answer in the backend. We may add a `total_amount_fiat` and a `total_fees_fiat` properties.
Then in the front-end, we just recover the 2 properties above and display them.

PR #506, opened 2020-10-16
Display fiat equivalents when sending coins

SylEze:
Solves #504 

Old behavior:
send_coins_without_fiat_eqivalent

New behavior:
send_coins_with_fiat_equivalent


src/atomicdex/pages/qt.wallet.page.cpp
```diff
@@ -270,6 +287,19 @@ namespace atomic_dex
                 {
                     j_out["withdraw_answer"]["fee_details"]["amount"] = j_out["withdraw_answer"]["fee_details"]["miner_fee"];
                 }
+                
+                // Add fees amount in fiat currency.
+                auto fee = j_out["withdraw_answer"]["fee_details"]["amount"];
```
Milerius, line 292: Spurious automatic conversion to `std::string` will rather prefer to use `get<std::string>()` to be sure.

naezith:
Seems good to me!

tonymorony (changes requested):
looks like USD prices calculating not correct in some cases (was my first test to broadcast KMD)
unfortunately atm 1 KMD cost a bit less than 41$ :)

image
image

SylEze:
> looks like USD prices calculating not correct in some cases (was my first test to broadcast KMD)
> unfortunately atm 1 KMD cost a bit less than 41$ :)
> 
> image
> image

I will check the function which does the conversion in the backend and the one which formats the result in the front-end

SylEze:
@tonymorony Can you try again with the last pull and tell me if it's working now please ?

tonymorony (approved):
lgtm now, thanks!

PR #516, opened 2020-10-20
Turkish translation

naezith:
It's not complete. Requires more contribution


tonymorony:
looks like it contains https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/515 changes also. do we really need both PRs?

naezith:
> looks like it contains #515 changes also. do we really need both PRs?

No, we don't.

tonymorony (approved):

Issue #521, opened 2020-10-22
[FEATURE REQUEST]: Display tooltip when hovering the band oracle icon


SylEze:
**Describe the solution you'd like**
When we hover the band icon with the mouse, we want to show a tooltip which tells from where the price is calculated.
Of course, adding a short delay (0.5s) when hovering the button is required, in order to do not annoy users when they pass on it.

**Additional context**
Demonstration:
![Display tooltip when hovering the band oracle icon](https://user-images.githubusercontent.com/14373103/96857352-2cbd8a80-145f-11eb-87c8-dd3dc4be37ab.gif)

Milerius:
closed by #522

PR #522, opened 2020-10-22
feat(band_icon_tooltip): Hovering band icon now displays tooltip

SylEze:
Tooltip content is "Price oracle powered by Band Protocol"

Solves #521

### DIFF
--- a/atomic_defi_design/qml/Components/CexInfoTrigger.qml
+++ b/atomic_defi_design/qml/Components/CexInfoTrigger.qml
@@ -16,7 +16,7 @@ DefaultMouseArea {
         delay: 500
 
         contentItem: DefaultText {
-            text_value: "Price oracle powered by Band Protocol"
+            text_value: qsTr("Price oracle powered by Band Protocol")
         }
     }
 }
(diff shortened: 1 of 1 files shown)


atomic_defi_design/qml/Components/CexInfoTrigger.qml
```diff
@@ -5,6 +5,18 @@ import QtQuick.Controls 2.15
...
     onClicked: cex_rates_modal.open()
+    hoverEnabled: true
+
+    DefaultTooltip {
+        visible: mouse_area.containsMouse
+
+        delay: 500
+
+        contentItem: DefaultText {
+            text_value: "Price oracle powered by Band Protocol"
```
Milerius, line 19: Usage of https://doc.qt.io/qt-5/qtquick-internationalization.html#1-use-qstr-for-all-literal-user-interface-strings is required here i think.

naezith: Yes, please use `qsTr(...)` so the text can be translated.

naezith (commented):
All good except the `qsTr(...)`.

smk762 (approved):
sighted in https://github.com/KomodoPlatform/atomicDEX-Desktop/actions/runs/321995166

Issue #523, opened 2020-10-22
[FEATURE REQUEST]: Display fiat equivalent in order line


SylEze:
**Describe the solution you'd like**
Add the price equivalent in fiat after the prices in coin.

**Additional context**
Result:
<img width="1444" alt="Capture d’écran 2020-10-22 à 19 52 45" src="https://user-images.githubusercontent.com/14373103/96912144-2d2b4500-14a2-11eb-94bd-905ac2471e6c.png">
<img width="1427" alt="Capture d’écran 2020-10-22 à 19 55 40" src="https://user-images.githubusercontent.com/14373103/96912152-2ef50880-14a2-11eb-84c4-0cae7efd6407.png">

Milerius:
closed by #524

PR #524, opened 2020-10-22
feat(order_line): Display fiat equivalents of base coin and rel coin in order line

SylEze:
Solves #523


naezith:
QML side looks correct but there is a problem in the backend I think, values are 0.
 image

tonymorony (approved):

PR #526, opened 2020-10-23
Fix electrum error index

naezith:


tonymorony (approved):
fixed issue on my side, thanks for a prompt fix

PR #528, opened 2020-10-24
feat(custom_coins): If there is a mismatch between coin cfg and our c…

Milerius:
…fg coin remove custom coin that cannot be enabled


smk762 (approved):
Sighted as working on https://github.com/KomodoPlatform/atomicDEX-Desktop/actions/runs/326233767

PR #530, opened 2020-10-25
Export CSV Polish

naezith:
- It was saving CSV by default anyways but now it also shows it at bottom:
image

- Also now it opens the save folder instead of default export folder.


smk762 (approved):
Works as expected, thanks :+1:

PR #531, opened 2020-10-26
fix osx bundling

Milerius:


smk762 (approved):
Trading view charts visible again :+1:

PR #540, opened 2020-10-29
Fix combobox was changing when list changes at 5.15.1

naezith:
Qt changed ComboBox functionality at 5.15.1, however value changes when list is updated and the index stays the same.

Fixing this:
image

If index changes, we know it's user input. 

So,
- if index changes, then value changes, we allow this change.
- if index does not change, but value changes, then we change the index to the old ticker, because we don't want value to change.

### DIFF
--- a/atomic_defi_design/qml/Exchange/Trade/DexComboBox.qml
+++ b/atomic_defi_design/qml/Exchange/Trade/DexComboBox.qml
@@ -105,11 +105,11 @@ DefaultComboBox {
 
                 Keys.onPressed: {
                     if(event.key === Qt.Key_Return) {
-                        if(ticker_list.length > 0) {
+                        if(control.count > 0) {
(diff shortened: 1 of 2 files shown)


smk762 (approved):
Confirmed working as expected

Milerius:
Don't merge until the bundle is fixed, @smk762 can you investigate a bit ? the problem seems to come from the CI

Milerius (changes requested):
due to https://bugreports.qt.io/plugins/servlet/mobile#issue/QTBUG-86759

naezith:
Please test the PR at 5.15.0 before reverting.

Milerius:
https://wiki.qt.io/Qt_5.15_Release 5.15.2 release tommorow, let's cross the finger :crossed_fingers: @naezith

naezith:
Qt gods help us.

Issue #545, opened 2020-10-30
[BUG]: Swap amounts in current currency is not updated when changing of currency


tonymorony:
**Describe the bug**
Swap value indication is incorrect when global curreny changing to crypto 

**To Reproduce**
1. Start swap with real value coin
2. Change global currency to non-fiat

![USD](https://user-images.githubusercontent.com/24797699/97709059-55dbbc00-1aec-11eb-9b4f-857952fcabb9.PNG)
![KMD](https://user-images.githubusercontent.com/24797699/97709071-583e1600-1aec-11eb-8b33-eb490b835529.PNG)
![BTC](https://user-images.githubusercontent.com/24797699/97709087-5e33f700-1aec-11eb-91ac-c6fde15ea789.PNG)

SylEze:
This also occurs when you switch from a fiat to another fiat

Milerius:
works for fiat @SylEze , need to implement for alternative currency too (KMD/BTC)

Milerius:
Can you test / close on last dev if everything is fine @smk762  ?

smk762:
![image](https://user-images.githubusercontent.com/35845239/105286140-b280db80-5bf0-11eb-84b2-8512433cc830.png)
![image](https://user-images.githubusercontent.com/35845239/105286184-c6c4d880-5bf0-11eb-8950-3807e5f43e63.png)
![image](https://user-images.githubusercontent.com/35845239/105286226-d5ab8b00-5bf0-11eb-9970-429f3a69814d.png)

Remains an issue in build 161b8c6. 
Values update elsewhere, but not in DEX > Orders page. Currency prefix updates, value of selected currency does not.

To reproduce, start a trade, return to dashboard page and click on wallet value (at the top) to toggle between Fiat / BTC / KMD value. Return to DEX > Orders page. Currency prefix updates, numeric value does not.

smk762:
Fixed in https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/635

PR #546, opened 2020-10-30
fix(orders_model): amounts values were not updated when changing currency

SylEze:
Solves #545

### DIFF
--- a/src/atomicdex/models/qt.orders.model.cpp
+++ b/src/atomicdex/models/qt.orders.model.cpp
@@ -21,11 +21,11 @@
 #include <boost/algorithm/string/case_conv.hpp>
 
 //! Project
-#include "atomicdex/services/mm2/mm2.service.hpp"
 #include "qt.orders.model.hpp"
 #include "atomicdex/events/qt.events.hpp"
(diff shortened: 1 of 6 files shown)


src/atomicdex/models/qt.orders.model.cpp
```diff
@@ -645,6 +610,48 @@ namespace atomic_dex
...
+        {
+            return determine_amounts_in_current_currency(contents.maker_coin, contents.maker_amount, contents.taker_coin, contents.taker_amount);
+        }
+        return determine_amounts_in_current_currency(contents.taker_coin, contents.taker_amount, contents.maker_coin, contents.maker_amount);
+    }
+    
+    void orders_model::on_current_currency_changed(const current_currency_changed&) noexcept
+    {
+        spdlog::trace("updated orders model");
+        refresh_or_insert_orders();
```
Milerius, line 652: I think here it's better to retrieve mm2 system and call the batch that refresh orders and swap since it's handling different internals for you

smk762 (approved):
Sighted in https://github.com/KomodoPlatform/atomicDEX-Desktop/actions/runs/339072245
working as expected

PR #549, opened 2020-11-01
feat(build_system): welcome unity build

Milerius:


smk762 (approved):
MultiOS loading and general usage tests working as expected.

PR #562, opened 2020-11-05
feat(maker_orders): update maker order in live

Milerius:


Milerius:
Live update maker order

smk762 (approved):
Sighted in https://github.com/KomodoPlatform/atomicDEX-Desktop/actions/runs/347031811
Note icons are still halloween themed, so order of merge is likely important.

PR #566, opened 2020-11-07
Fix CheckBoxes become black after entering input to the search at Enable Coin Modal

naezith:
We are using the Universal theme, the colors are set in a separate file, independent from the QML code. 

Since I was implementing the light theme, I removed those static color codes. Defined them in the top level QML code, at the application window.

```qml
    Universal.theme: Style.dark_theme ? Universal.Dark : Universal.Light
    Universal.accent: Style.colorQtThemeAccent
    Universal.foreground: Style.colorQtThemeForeground
    Universal.background: Style.colorQtThemeBackground
```

It worked fine. However at the Enable Coin Modal, if you filter the coins list by entering text in the search bar, all the checkboxes and their texts become black, means that theme colors are ignored.

I solved it by copying those same lines to my custom wrapper components of stock Qt components such as `CheckBox`.

So we better test it by interacting the application, seems like Qt has a bug somehow. Colors work at first launch but breaks after interaction. It should be fine now.

### DIFF
--- a/atomic_defi_design/assets/languages/atomic_defi_en.ts
+++ b/atomic_defi_design/assets/languages/atomic_defi_en.ts
@@ -2275,11 +2275,6 @@
         <source>Privacy</source>
         <translation type="unfinished"></translation>
     </message>
-    <message>
-        <location filename="../../qml/Sidebar/SidebarBottom.qml" line="40"/>
-        <source>Light UI</source>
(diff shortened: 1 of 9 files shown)


smk762 (approved):
Sighted in https://github.com/KomodoPlatform/atomicDEX-Desktop/actions/runs/351308588
and confirmed working as expected on all OS

PR #576, opened 2020-11-16
feat(osx): new artifacts (installer)

Milerius:


smk762 (approved):
mac 7z extracts to installer, installer works, app works, and uninstaller (maintenance tool) works.

PR #603, opened 2020-12-04
update mm2.1 RC

tonymorony:


smk762 (approved):
Basic mm2 ops on all OS tested without issue

PR #610, opened 2020-12-08
Added GETH to default coins

theblackmallard:
GETH is Guarded Ether. Since ETH2.0 staking rewards cannot be withdrawn at this stage Guarda wallet recently implemented of a 1:1 ETH token that is specifically paid out to their users who are within their Staking pool. So this can be withdrawn from Staking rewards and when withdrawing is enabled, can be used to claim ETH from the pool at 1:1 ratio.


smk762:
stale PR, please resubmit to latest version's coins file if still required.

PR #635, opened 2021-01-21
fix(swap_amount): update fiat value also for orders

Milerius:


smk762 (approved):
Confirmed working as expected.

PR #638, opened 2021-01-22
fix(addressbook): Add better alignment of UI

SylEze:


smk762:
Text alignment still looks a bit high, buttons have disappeared!
image

PR #663, opened 2021-02-01
feat(exporter_service): add exporter service

Milerius:


smk762:
Looks good, respects the filter.
On linux no ".csv" in filename, other OS ok.
Only FR I could add is to limit filter dropdown to coins within history rather than all.

Milerius:
> Looks good, respects the filter.

> On linux no ".csv" in filename, other OS ok.

Should be fixed in https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/663/commits/47a54732a4b8daf3447b769da8877afab0a0eab9

> Only FR I could add is to limit filter dropdown to coins within history rather than all.

I agree with you, we would need a modal before choosing the name of the file which would allow us to choose between exporting the total history or only the current page, we can do it at the next split.

smk762 (approved):
CSV extension now present in linux

PR #665, opened 2021-02-03
feat(addressbook): Addressbook tweaks

### LINKED ISSUE #645: [FEATURE REQUEST]: Addressbook tweaks
A few suggestions to improve Addressbook UX - sometimes it feels a bit like I'm clicking too much to get where I want to be.

On this modal, I think it would be better to show the complete list of stored addresses for a contact in the table, rather than have to define the coin or coin category to have them displayed. 

image

IMO, the "change of wallet type" button would be better as a filter than a selector (e.g. show all by default, filter to selected)

This change would mean that when user is adding a new address, they would need to click the "new" button first, then select the coin. Currently the "new" button is inactive/disabled until the "change of wallet type" is defined.

Another feature that could reduce number of clicks to get where I'm going would be to auto-apply the filter of addresess when clicking into addressbook from the wallet page, to match the coin active on the wallet page's send modal.

image

Currently if I click on the "Address book" button in this modal, it is no different than clicking on to addressbook tab from the sidebar menu, then I have to specify the coin all over again.

The addressbook selection flow is "contact > coin > address". Perhaps it might be more efficient to have the addressbook table display ALL contacts, coins, and addresses, and then allow filtering by contact, filtering by coin and/or filtering by address. 
i.e. separate the "add new" flow from the "select to send" flow.

Rough concept image below
image

Note edit/delete icons added to row for each address for future removal or modification.

Form to add/edit contacts like below.
image

I think something like this could streamline the use of the addressbook, allowing for less clicks to achieve the same outcomes.

SylEze:
Fixes #645


smk762:
- buttons/layout now appear aligned 
- no mispellings of "address" sighted.

A couple more tweaks:
- Currently when adding a new address, in the "select wallet type" modal, when entering something into search it does not expand the menu tree. Suggest either tree expanded by default, or expands once 2 char entered.
- Clicking the "+" button below Tags heading, then adding text in the "add a new tag" modal, clicking "Add" does nothing, and I have to cancel out. App log returns `[main.prerequisites.hpp:120]: qrc:/atomic_defi_design/qml/Wallet/AddressBookNewContactCategoryModal.qml:0: ReferenceError: modelData is not defined`

smk762:
Confirm tag button now working as expected.

Milerius:
What's missing to be merged @smk762 @SylEze ?

src/atomicdex/models/qt.addressbook.contact.model.cpp
```diff
@@ -169,31 +210,52 @@ namespace atomic_dex
...
+            const auto& coins_type_list        = portfolio_pg.get_global_cfg()->get_all_coin_types();
+            const auto  create_address_entries = [&](const QString& type)
+            {
+                if (!addrbook_manager.has_wallet_info(m_name.toStdString(), type.toStdString()))
+                {
+                    return;
+                }
+                const auto& addresses = addrbook_manager.get_wallet_info(m_name.toStdString(), type.toStdString()).at("addresses");
+    
+                for (auto it = addresses.begin(); it != addresses.end(); ++it)
```
Milerius, line 250: any reason why not using const iterator ? `cbegin()` `cend()`

smk762:
Just the FR - `Currently when adding a new address, in the "select wallet type" modal, when entering something into search it does not expand the menu tree. Suggest either tree expanded by default, or expands once 2 char entered.`

Milerius:
> Just the FR - `Currently when adding a new address, in the "select wallet type" modal, when entering something into search it does not expand the menu tree. Suggest either tree expanded by default, or expands once 2 char entered.`

ok once @SylEze will push the fixes / enhancement we can merge this one and then 0.4.0

smk762 (approved):
Tested and working as expected.

PR #669, opened 2021-02-09
feat(coingecko_provider): start the coingecko provider

Milerius:
Remove wallet charts since coingecko doesn't provide the same data format


smk762 (dismissed):
"enable all" functioning without problem.
NaN on priceless pairs resolved.
null recent swap handled.
LGTM.

Issue #420, opened 2020-08-10
[FEATURE REQUEST]: Sort fiat option menu selections


smk762:
Simple feature request: Currently Fiat options list is unsorted. Can this be made alphabetical (maybe with major currencies at top for convenience, followed by others)
![image](https://user-images.githubusercontent.com/35845239/89788767-3cad1a00-db52-11ea-8a0b-22b052eca36d.png)

Coingecko options are an example of the major/other option split - 
![image](https://user-images.githubusercontent.com/35845239/89788862-5babac00-db52-11ea-8e13-773fb94656f7.png)

Adding full currency name might look nice also, but probably people will know what they are looking for anyway without it.

naezith:
@Milerius we need to do it together, with the `textRole`, `valueRole` stuff again

cipig:
would be nice if the list of the coins to be enabled (when clicking on +) would be sorted too, alphabetically by coin symbol or name

smk762:
sighted and working as expected, then merged.

PR #671, opened 2021-02-10
feat(settings): Enhance fiats selection in settings menu

SylEze:
Solves #420 

Features:
 - Sorts fiats combo box
 - Display 6 suggested fiats below the fiats combo box


smk762 (approved):
Sighted and testing - works as expected.

Issue #696, opened 2021-02-17
[BUG]: Send page contact list does not display contacts with coin type addresses (e.g. ERC20)


SylEze:
**Describe the bug**
The send page contact list is displaying only contacts with an address type equivalent to the selected coin. 

However let say that the given coin is a `Smart Chain` coin: KMD, the contact list will not display any address of type `Smart Chain`, only the KMD ones

**To Reproduce**
- Having a contact with a `Smart Chain` address
- Go to the wallet page
- Choose `KMD` and click on `Send`
- Click on `Address Book`
- The contact with a `Smart Chain` address does not appear even though it should be compatible with `KMD`

**Expected behavior**
Display every contact which has a compatible address with the selected coin.

Milerius:
done in #697

PR #697, opened 2021-02-17
fix(send): Display contacts with coin type addresses in contact list

SylEze:
Solves #696


smk762 (approved):
Tested with specific coin and coin category, looks good.

Issue #721, opened 2021-02-26
[FEATURE REQUEST]: Sort the My Wallets section alphabetically.


lightspeed393:
**Is your feature request related to a problem? Please describe.**
The wallets in the `My Wallets` section when logging in are in a random order.

**Describe the solution you'd like**
The wallets are listed alphabetically.

Milerius:
Implemented in #722

PR #722, opened 2021-02-27
feat(wallet): sort wallets name in get_wallets

Milerius:
Implement #721


smk762 (approved):
Wallet accounts on login now alphabetically sorted.

PR #730, opened 2021-03-03
add coingecko guide GIF

cipig:


Milerius:
@smk762 custom coins addition was broken due to change from coinpaprika to coingecko, can you test to check if this work ? 

I successfully added https://www.coingecko.com/fr/pi%C3%A8ces/nexo on my side

smk762 (approved):
Tested on Win / Linux with no problems

PR #773, opened 2021-03-27
Portfolio improvements

ssakone:


Milerius:
Great work @ssakone 

Few improvements needs to be achieved before we can merge this pr:

ETH:
image

And sometimes we have this effect:
image

qml.qrc
```diff
@@ -448,5 +448,7 @@
         <file>atomic_defi_design/qml/Wallet/Wallet.qml</file>
         <file>atomic_defi_design/qml/main.qml</file>
         <file>qtquickcontrols2.conf</file>
+        <file>atomic_defi_design/qml/Portfolio/PieItem.qml</file>
+        <file>atomic_defi_design/qml/Components/DexFadebehavior.qml</file>
```
Milerius, line 452: qrc file is now sorted, please respect the sort order

tonymorony:
great progress @ssakone !

I think it might be worth adding also portfolio value movement chart but we have to think about the best/optimal way on getting user balance and crypto price retrospectively - probably it'll require a separate microservice 🤔

Screenshot 2021-03-29 at 17 46 47

qml.qrc
ssakone: Qt creator add dynamicaly new line in qrc

Milerius:
Thanks you for the awesome last push @ssakone few remarks left on this pr:

- The mouse doesn't show the value yet when pointing specific part of the chart:

- In some places of the chart the part of the gradient exceeds the line:

Capture d’écran 2021-04-05 à 05 53 24

- The date always show the same day:

Capture d’écran 2021-04-05 à 05 54 46

- The pie chart always show $ sign despite i'm moving the mouse:

https://user-images.githubusercontent.com/21139416/113535378-865fde80-95d3-11eb-801c-ff59069ce513.mov

- The chart legends  are not update each time i disable / enable a coin, the chart also exceed the limit in this case:

https://user-images.githubusercontent.com/21139416/113535620-36354c00-95d4-11eb-9e3a-45f14d08cc22.mov

- Sometimes chart are empty:

Capture d’écran 2021-04-05 à 05 56 59

(if i switch page and comeback the chart load)

- There is no spinner for the chart that explain if the chart is busy loading or not

- There is no way to choose the timeframe of the chart yet:

Capture d’écran 2021-04-05 à 05 59 14

tonymorony:
`Reduce a bit the size of the chart / pie chart move the add assets button near the search bar so we can see a bit more the bottom array of coins.`

yeah, also think that it's too huge now and that assets adding button is logically disconnected from assets list

dates axe not working atm it seems

text on assets adding a button is not centered:
Screenshot 2021-04-05 at 17 53 05

by some reason graph drawn to 0:

Screenshot 2021-04-05 at 17 49 35

it looks like the right axis scale is random:

Screenshot 2021-04-05 at 17 59 04

would be great to scale it linear with rounded step, e.g. :

8A94D9D0-33F5-4863-953F-A2E06493D714

also, it would be great to display the portfolio change amount for a selected timeframe as on the screenshot above (you can check Blockfolio mobile app for reference)

Milerius:
@ssakone Excellent improvements in your last commits, there is my new review:

- [ ]  The mouse doesn't show the value yet when pointing specific part of the chart

- [ ] `In some places of the chart the part of the gradient exceeds the line` - Now we have line, we should probably use line by default and disable spline since it's possible to fix this problem otherwise

- [ ] It seems that there is a visible latency time between the moment when the chart is no longer busy and when the chart is redrawn with the modified axes:


https://user-images.githubusercontent.com/21139416/113669964-c8ad1c80-96b4-11eb-9e48-fe2d28c50210.mov

Please use a spinner again if the chart is not finished being drawn, this way once the spinner is finished - the user is guaranteed to have the final chart with the correct values.

Milerius:
Created separate issue for stuff that are not finished

PR #802, opened 2021-04-15
fix(portfolio): fix bug (portfolio page)

ssakone:


tonymorony:
`+` disappeared for me from the button label:

image

cipig (approved):
looks good, thanks for the fix
image

PR #805, opened 2021-04-15
feat(header): custom header bar

Milerius:
resize/maximized need to be tested on all platform @smk762


smk762:
Issue https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/798 persists in https://github.com/KomodoPlatform/atomicDEX-Desktop/actions/runs/751069316
image

can be resolved by manually extending page down. Resizing like this there is a slight lag before buttons and input fields are usable.

Standard minimize / maximize / close buttons working as expected.

smk762:
Same issue persists in win/mac when app window reduced to smallest size.
Standard minimize / maximize / close buttons working as expected on Windows.
Standard maximize / close buttons working as expected on Mac, but minimise button does nothing, and am unable to manually resize by dragging from edges.

PR #808, opened 2021-04-15
Fix portfolio bug

ssakone:


cipig:
image
the 1M chart is lacking the last 7 days, it goes directly from 07.04. to 15.04.

Milerius:
> image
> the 1M chart is lacking the last 7 days, it goes directly from 07.04. to 15.04.

This is a problem from coingecko API not related to us unfortunately

smk762 (approved):

PR #836, opened 2021-04-20
Theme light improve

ssakone:


cipig:
wrong branch? btw, why not always enable KMD? KMD, BTC, DOGE for DogeDEX... could be needed later: https://github.com/KomodoPlatform/atomicDEX-API/issues/715

PR #841, opened 2021-04-21
fix(qml): Addressbook table style inconsistent with rest of app

SylEze:
Should fix https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/823


Milerius:
Can you test it please @smk762 ?

smk762 (approved):
Looking good now, dont see any white border anymore

PR #842, opened 2021-04-22
fix(text_overflow): reduce little comboBox second text size

ssakone:


smk762 (approved):
Overflow value issue appears to be resolved, and extra space between best price cols sighted

PR #848, opened 2021-04-22
feat(lang): new qml approach

Milerius:
Lang change should act exactly as before - no crash, we use QSettings now


smk762 (approved):
changing lang on all OS as expected without issue.

PR #859, opened 2021-04-25
feat(app): Add option to enable 2FA

SylEze:
Option to enable 2FA is located in: Settings -> Security

Can you recheck that 2FA is still working correctly on both Windows/MacOS please ?


atomic_defi_design/qml/Settings/SettingModal.qml
```diff
@@ -323,6 +323,31 @@ Qaterial.Dialog {
                                 sourceComponent: CamouflagePasswordModal {}
                             }
 
+                            // Enabled 2FA option. (Disabled on Linux since the feature is not available on this platform yet)
+                            RowLayout {
+                                enabled: API.qt_utilities.get_current_os() !== "LINUX" // Disable for Linux.
```
Milerius, line 328: use Qt.os ?

smk762:
Windows did not ask for password after setting flag, also tested after restarting, still no password prompt.

Working as expected on Mac.

Milerius:
@smk762 can you open a separate issue for windows problem, i guess it's not a blocker, we can just solve it into another PR

PR #872, opened 2021-04-30
feat(notification_modal): Change notification display

ssakone:
- New Notification display Box
- Fix duplicate balance change message (NOT IMPLEMENTED)
- Save notification in db (REMOVED)

*Question*
1. If I understood correctly the duplication of the notification change of balance comes with two notifications sent and received, now which is the right one, the first or the second?
2. I m not able to save data in localStorage, i go this error `QtQuick.LocalStorage not installed` maybe its required some thing, i can use local sqlite but what location exactly i can stored it. 

image


tonymorony:
I continue to see notifications if I disable it in settings:

image

----

```
If I understood correctly the duplication of the notification change of balance comes with two notifications sent and received, now which is the right one, the first or the second?
```

it might be different - sometimes it's send/receive, sometimes it receives/send (as I do understand it's related to the mempool specifics and behave like that for UTXO coins only)

In the end, what is important is balance change: maybe for now if we noticed balance change we should not display notification instantly and wait let's say 10 seconds, and if during these 10 seconds there were other changes display it as a single notification to prevent such false alarms. Maybe this trick will help us to reduce such false alarm notifications amount.

Afaik @Milerius reflected on this problem before and got some thoughts on it.

----

`I m not able to save data in localStorage, i go this error QtQuick.LocalStorage not installed maybe its required some thing, i can use local sqlite but what location exactly i can stored it.`

Let's not store notifications locally for now, visual improvement is fine enough

----

Converted PR to the draft since there are some commented code blocks.

tonymorony:
hmm not starting on my mac:

https://paste.ubuntu.com/p/NKZtN2bvs9/

smk762:
Mac failing to start here also with same errors
```[16:26:44] [info] [internet.checker.service.cpp:131] [3137634]: Connectivity is true for the endpoint: https://api.coinpaprika.com/
[16:26:44] [info] [ip.checker.service.cpp:90] [3137697]: my ip address is: [101.185.130.123]
[16:26:45] [warning] [main.prerequisites.hpp:125] [3137585]: QQmlApplicationEngine failed to load component
[16:26:45] [warning] [main.prerequisites.hpp:125] [3137585]: qrc:/atomic_defi_design/qml/Components/DexWindowControl.qml:7:1: "Screens": no such directory
[16:26:45] [warning] [main.prerequisites.hpp:125] [3137585]: qrc:/atomic_defi_design/qml/main.qml:53:5: Type DexWindowControl unavailable
[16:26:45] [error] [main.prerequisites.hpp:139] [3137585]: sigabort received, cleaning mm2
No matching processes belonging to you were found
[16:26:45] [error] [main.prerequisites.hpp:145] [3137585]: stacktrace:  0# mm2::api::from_json(nlohmann::basic_json<std::__1::map, std::__1::vector, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, bool, long long, unsigned long long, double, std::__1::allocator, nlohmann::adl_serializer, std::__1::vector<unsigned char, std::__1::allocator<unsigned char> > > const&, mm2::api::fraction&) in /Applications/atomicdex-desktop.app/Contents/MacOS/atomicdex-desktop
 1# _sigtramp in /usr/lib/system/libsystem_platform.dylib
 2# QWindow::winId() const in /Applications/atomicdex-desktop.app/Contents/Frameworks/QtGui.framework/Versions/5/QtGui
 3# main in /Applications/atomicdex-desktop.app/Contents/MacOS/atomicdex-desktop
 4# main in /Applications/atomicdex-desktop.app/Contents/MacOS/atomicdex-desktop
```

smk762:
The vout index might help with determining a send / receive in transactions where the vin address also appears as a vout.
For example see: https://kmd.explorer.dexstats.info/insight-api-komodo/tx/92e217979d934dd66481e4828a803119ac63c8958d5e1e6bfe1e4838be7f8da9

Vin address = RTh4Ha1MUfg6CQ69mbb5d9ZwuZDYe4NRYK (this is the address funds came from)

Vout address index 0 is the destination address: bWg1wLsXD3JGbFFwwT8eegaMvXKvWP4nMK

Vout index 1 has no destination address - but it has an OP_RETURN (likely this is a dex fee transaction)

Vout index 2 has same address as Vout: this indicates the change returning from the transaction.

Milerius:
> The vout index might help with determining a send / receive in transactions where the vin address also appears as a vout.
> For example see: https://kmd.explorer.dexstats.info/insight-api-komodo/tx/92e217979d934dd66481e4828a803119ac63c8958d5e1e6bfe1e4838be7f8da9
> 
> Vin address = RTh4Ha1MUfg6CQ69mbb5d9ZwuZDYe4NRYK (this is the address funds came from)
> 
> Vout address index 0 is the destination address: bWg1wLsXD3JGbFFwwT8eegaMvXKvWP4nMK
> 
> Vout index 1 has no destination address - but it has an OP_RETURN (likely this is a dex fee transaction)
> 
> Vout index 2 has same address as Vout: this indicates the change returning from the transaction.

The problem is that the notification for the transactions is based on the change of balance, not on the study of the transaction history - for performance reasons we cannot track the transaction history of all the assets. enabled, however with the upcoming arrival of websocket this kind of thing should no longer be a problem. What is needed is a light way of detecting new transactions without having to check the transaction history of all assets.

ssakone:
> 
> 
> hmm not starting on my mac:
> 
> https://paste.ubuntu.com/p/NKZtN2bvs9/

Fixed and checked on Mac.

smk762:
Confirm also working here on mac with https://github.com/KomodoPlatform/atomicDEX-Desktop/actions/runs/806795583

Will do multiOS tests this evening and approve after meeting.

atomic_defi_design/qml/Components/DexModel.qml
```diff
@@ -0,0 +1,61 @@
+import QtQuick 2.15
+import QtQuick.LocalStorage 2.15
+
+ListModel {
+    id: control
+    property string dbName: "Test00000"
```
tonymorony, line 6: it looks like this model not using atm, do you plan to utilize it later?

Milerius: I agree - this file need to be removed, on top of that the name `DexModel.qml` seems not appropriate for a storage model, something like `NotificationStorageModel.qml` would be more appropriate, if it's not used you can simply delete it

atomic_defi_design/qml/Components/DexWindow.qml
```diff
@@ -9,4 +9,6 @@ import "../Constants"
 ApplicationWindow {
     id: window
     flags: Qt.FramelessWindowHint | Qt.Window
+
```
Milerius, line 12: Spurious space, can be removed

qml.qrc
```diff
@@ -469,5 +469,7 @@
         <file>atomic_defi_design/qml/Portfolio/AmountChart.qml</file>
         <file>atomic_defi_design/qml/Portfolio/AssetPieChart.qml</file>
         <file>atomic_defi_design/qml/Portfolio/TableDex.qml</file>
+        <file>atomic_defi_design/qml/Components/DexModel.qml</file>
```
Milerius, line 472: DexModel.qml can be removed, please ensure also that the qml.qrc is sorted

atomic_defi_design/qml/Dashboard/NotificationsModal.qml
```diff
@@ -264,121 +293,103 @@ BasicModal {
...
+                    }
+                    height: _column.height+10
+                    width: list.width
+                    property int stat: {
+                        if (notifications_list[index].title.indexOf("You received")!==-1 ) {
+                            return 0
+                        } else if(notifications_list[index].title.indexOf("You sent")!==-1) {
+                            return 1
+                        } else {
+                            return 2
```
Milerius, line 314: This is not a good way to detect notification, according to the old code, we was able to know which kind of notification are received and how to treat them.

- What means `stat` field ? 
- Later in the code you are using this `stat` field to get others infos such as color or icon, this is a violation of https://en.wikipedia.org/wiki/Don%27t_repeat_yourself (DRY rules) - you can do a single function that will set every information you need once, you can also at the moment you receive the notification set it in a way that retrieving information will be easier such as a map:

`notifications_map[notifications_list[index].kind].color`
`notifications_map[notifications_list[index].kind].icon`

Where `kind` could be :

```cpp
enum notification_kind
{
   send,
   receive,
   others
}
```

smk762:
minimise button in OSX appears to not be working (fine on win/linux)

Milerius:
the code looks good to me now, @tonymorony @smk762 can you test again and approve if everything work as expected ?

Milerius (approved):

smk762 (approved):
Looking odd here, I like the "click on item goes to screen where thing happened" functionality. 
Tested on all OS, with and without notification display enabled - all good.

PR #877, opened 2021-05-02
add Diminutive Coin (DIMI)

cipig:


smk762:
DIMI sighted in "Enable assets" list, and enables without problem (after resetting assets config).

Two entries for SmartFi in "enable list, with differing logos:
image

LYNX not displayed in "Enable assets" list, even after resetting assets config.

cipig:
LYNX needs a new build. since it was made before coins file was updated
confused about Smartthingy too, will rename the UTXO coin to SmartUSD, like in coins file

smk762 (approved):
I can see LYNX now, still see DIMI, and SmartFi/SmartUSD are distinctly displayed

Issue #884, opened 2021-05-05
[FEATURE REQUEST]: Limit simultaneously activated coins amount 


tonymorony:
cf: https://github.com/KomodoPlatform/AtomicDEX-mobile/issues/745

if the user activating all availiable coins now it might break the app (until he "reset assets configuration") with a high probability 
also, it might create a lot of file descriptors cf: https://github.com/KomodoPlatform/atomicDEX-API/issues/791 what can cause huge loading on user's machine/network and infra

I do propose 50 simultaneously activated coins as a starting point / reasonable limit

Milerius:
> cf: [KomodoPlatform/AtomicDEX-mobile#745](https://github.com/KomodoPlatform/AtomicDEX-mobile/issues/745)
> 
> if the user activating all availiable coins now it might break the app (until he "reset assets configuration") with a high probability
> also, it might create a lot of file descriptors cf: [KomodoPlatform/atomicDEX-API#791](https://github.com/KomodoPlatform/atomicDEX-API/issues/791) what can cause huge loading on user's machine/network and infra
> 
> I do propose 50 simultaneously activated coins as a starting point / reasonable limit

Maybe a child issue of https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/843

Milerius:
implemented in #886

PR #886, opened 2021-05-05
feat(coins): add a settings to allow limitation of simultaneous coin enabled

Milerius:
first step for #884 

cc @SylEze  or @ssakone 

You can simply use the size of the portfolio model size to get the number of coins -> https://github.com/KomodoPlatform/atomicDEX-Desktop/blob/a80c2a3c7ea68e952bd97125987b5971753bd45b/src/core/atomicdex/models/qt.portfolio.model.hpp#L44

You can then use the new settings, and the number of checked coins in the enable modal to ensure that the user cannot enabled more than `MaximumNbCoinsEnabled`

You can make a new visual settings for changing this number with predefined values: [50, 75, 100, 150, 200]

Please enhance the user experience of `select all` in the modal, popup/warn the user that you select only the N first entries in case limit is reached, add tag button to automatically see ERC20, BEP20, QRC20, SmartChains or UTXO coins

If select all of a filtered research is less than `MaximumNbCoinsEnabled - portfolio_model.length` you can keep the current behavior, you can also show with hints, how much user can select coins

You can directly push to this pull request

Some example:

- I have 10 coins enabled, MaximumNbCoinsEnabled is equal to 50, i'm able to enable 40 (limit - nb_enabled_coins) assets

Please also adapt the code from best orders that propose to the user to enable a coin, same check should also occured here, add a well explained message in case user want to enable a coin but limit is reached.

@smk762 if you can provide a good english message for a new FAQ entry regarding simultaneous coins activated limitation (bandwith, performance, enable coins that you are willing to use) - it's will be fantastic

cc @tonymorony @smk762


SylEze:
Steps:

- [ ] Visual settings to set the maximum number of enablable coins: [10, 20, 50, 75, 100, 150, 200]
- [ ] In enable coins modal, warn the user if he selects only the N first entries in case limit is reached
- [ ] Ensure `Send` button that requires a parent chain to be enabled, shows another modal than the enable one in case the limit is already reached. (e.g. https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/893)

atomic_defi_design/qml/Settings/SettingModal.qml
```diff
@@ -156,6 +156,21 @@ Qaterial.Dialog {
...
+                                width: parent.width-30
+                                anchors.horizontalCenter: parent.horizontalCenter
+                                height: 50
+                                DexLabel {
+                                    Layout.alignment: Qt.AlignVCenter
+                                    Layout.fillWidth: true
+                                    text: qsTr("Maximum number of enabled coins")
+                                }
+                                DexComboBox {
+                                    model: [10, 20, 50, 75, 100, 150, 200]
```
Milerius, line 169: Here it's should be possible to detect the maximum value, which is the global cfg size i believe

SylEze:
> @SylEze every time i restart the current selection of combobox is empty
> image

Should be fixed in 5b039130760d9808e8ee9a0ac882eb34017ea151

smk762:
Limit appears to be enforced as expected in standard usage, but if changing assets limit to a lower value, previously selected items remain selected, and attempting to activate from this state results in nothing being activated.

https://user-images.githubusercontent.com/35845239/117929918-f0d11080-b32f-11eb-8462-373d90acd113.mp4

Also suggest that the "Increase Assets Limit" button be renamed to "Change Assets Limit"

Milerius:
> Limit appears to be enforced as expected in standard usage, but if changing assets limit to a lower value, previously selected items remain selected, and attempting to activate from this state results in nothing being activated.
> 
>  enable_limit.mp4 
> Also suggest that the "Increase Assets Limit" button be renamed to "Change Assets Limit"

Thanks you for the report ! yeah i agree in general @SylEze we can probably unselect all in case the limit is changed on the fly ^^ !

Milerius:
> Limit appears to be enforced as expected in standard usage, but if changing assets limit to a lower value, previously selected items remain selected, and attempting to activate from this state results in nothing being activated.
> 
>  enable_limit.mp4 
> Also suggest that the "Increase Assets Limit" button be renamed to "Change Assets Limit"

Do you confirm that trying to enable coins from the send modal and best orders handle also the limit ?

SylEze:
> Limit appears to be enforced as expected in standard usage, but if changing assets limit to a lower value, previously selected items remain selected, and attempting to activate from this state results in nothing being activated.
> enable_limit.mp4
> 
> Also suggest that the "Increase Assets Limit" button be renamed to "Change Assets Limit"

Thanks for your report, I will patch it asap

SylEze:
> > Limit appears to be enforced as expected in standard usage, but if changing assets limit to a lower value, previously selected items remain selected, and attempting to activate from this state results in nothing being activated.
> > enable_limit.mp4
> > Also suggest that the "Increase Assets Limit" button be renamed to "Change Assets Limit"
> 
> Thanks for your report, I will patch it asap

Patched by 43157b05a610885e3c7dc777ed4b4ecb882c68c7

atomic_defi_design/qml/Wallet/EnableCoinModal.qml
```diff
@@ -125,8 +130,32 @@ BasicModal {
...
+                        .arg(setting_modal.enableable_coins_count - API.app.portfolio_pg.portfolio_mdl.length)
+                        .arg(coin_cfg_model.checked_nb)
+            }
+        }
+
         // Buttons
         footer: [
+            DefaultButton {
+                property var enableable_coins_count: setting_modal.enableable_coins_count;
+                text: qsTr("Increase assets limit")
```
smk762, line 154: Change this to "Change assets limit", as user might be decreasing it.

smk762 (approved):

atomic_defi_design/qml/Wallet/EnableCoinModal.qml
Milerius: implemented in https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/886/commits/2d32dcb85d16f8cf5ea3aece86a8710f49764646

PR #890, opened 2021-05-06
feat(custom_cfg): separate custom coins into a different cfg

Milerius:
need to be tested:

- Custom coins addition
- Will create a custom cfg for custom tokens eg: `/Users/milerius/Library/Application Support/AtomicDex Desktop/config/custom-tokens.Roman.json`
- Delete/Enable/Disable will modify value in `custom-tokens.walletname.json`
- Now when you reset asset cfg that doesn't delete your custom tokens anymore, can create a missmatch with mm2 cfg same as before, a simple restart solve the problem

To obtains a missmatch:

- Add a custom coin
- Restart
- Enable It
- Reset asset cfg
- Enable it again

Please ensure you also get the notification on #872 too (you can merge this branch into the one from ssakone locally)

### DIFF
--- a/atomic_defi_design/assets/languages/atomic_defi_en.ts
+++ b/atomic_defi_design/assets/languages/atomic_defi_en.ts
@@ -1352,6 +1352,18 @@
     </message>
     <message>
         <location filename="../../qml/Constants/General.qml" line="452"/>
+        <location filename="../../qml/Constants/General.qml" line="456"/>
+        <source>%1 parent chain is not enabled</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <location filename="../../qml/Constants/General.qml" line="454"/>
+        <location filename="../../qml/Constants/General.qml" line="458"/>
+        <source>%1 parent chain balance is 0, a non-zero balance is required to pay the gas</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <location filename="../../qml/Constants/General.qml" line="460"/>
         <source>Unknown Error</source>
         <translation type="unfinished"></translation>
     </message>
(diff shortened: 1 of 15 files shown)


Milerius:
i confirm @tonymorony @smk762 @ssakone new notification are absolutely broken in this regards -> 
Capture d’écran 2021-05-06 à 11 00 15

Please check the difference with https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/890/commits/882652d2ef1697348cf255a58e497a6a8f375d14

In the old notification modal we have a restart button, why this button is not present anymore, and why it's was not reported on #872  ?

Milerius:
Should be solved in https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/890/commits/e6bc229e92157afd839b20e0956cddd22f78ce0b

smk762:
Can confirm that custom tokens that were added will persist after a coin config reset.
Possible issue - If adding a custom token that already exists, it will create a duplicate entry.

image

Expected behaviour would be to check if coingeckoID exists within standard config, and then let user know coin is already available and not progress towards creation of a duplicate.
image

Also, if not too difficult, it would be nice to prettify the custom-tokens.json file (it was minified by default, I had to prettify)

smk762:
A mismatch restart button does appear in notifications, and on clicking, restarts the app as expected.
image

image

Milerius:
> A mismatch restart button does appear in notifications, and on clicking, restarts the app as expected.
> image
> 
> image

We check if a coin already exist in fact, the problem is that a recent upgrade add `-ERC20` yet, that's why you have this possibility, should not be possible with new users, other than that we need to check with @cipig that our current ERC20 coins have the `ERC-20` suffix addition

smk762 (approved):

PR #925, opened 2021-05-18
fix run time icon

ssakone:


atomic_defi_design/qml/main.qml
```diff
@@ -42,12 +42,29 @@ DexWindow {
...
+    }
+    DexBusyIndicator {
+        anchors.centerIn: parent
+        running: app_loader.status === Loader.Loading
+        visible: running
+    }
+    Shortcut {
+        sequence: "Ctrl+L"
+        onActivated: app_loader.reload()
+    }
```
Milerius, line 62: Please try to add it only when hot reload is enable: `engine.rootContext()->setContextProperty("debug_bar", QVariant(true));` if i remember you have that when hot reload is enabled

ssakone:
Sure

Le mar. 18 mai 2021 à 18:08, Sztergbaum Roman ***@***.***> a
écrit :

Milerius (approved):

PR #931, opened 2021-05-21
Update linux_script.sh

smk762:
change gcc version


smk762:
converted to draft while Github sorts out its ubuntu 18 env - https://github.com/actions/virtual-environments/issues/3432

PR #936, opened 2021-05-23
🍱 Update atomic_defi_tr.ts

semsdev:
Translated the fields in which translation is not completed into Turkish.


Milerius:
> Translated the fields in which translation is not completed into Turkish.

Thanks for the pr, it's seems that lot of error are present in the traduction according to linguist -> 

Capture d’écran 2021-05-23 à 14 45 18

Would you mind to open the ts file with linguist and use it step by step for the translation ?

Milerius (changes requested):
check comment on the pr

Milerius (approved):

PR #937, opened 2021-05-24
Update linux_script.sh

smk762:
see https://github.com/actions/virtual-environments/pull/3436


smk762:
no bueno - did not work

PR #945, opened 2021-05-27
feat(best_orders): best orders need to be changed each time the volum…

Milerius:
…e change

to test:

- [x] Changing the volume manually trigger best orders
- [x] Try to spam volume change to be sure no crash is occuring on any platform
- [x] CPU should spike a bit during this spam but goes done once every request is finished


smk762:
Best orders appears on dex page load. When changing volume via slider or form input, best order values change proportionately, and seem to be filtered out at lower volumes.

No problem with slider spamming on linux

smk762:
No crash on windows, but its a bit slower to update when spamming, and might blank out for a bit.

smk762:
Mac also no crash on slider spam

smk762 (approved):

PR #946, opened 2021-05-27
feat(swaps): update max_taker_vol when swap ongoing 2/3 in order to u…

Milerius:
…pdate the max taker vol

to test:

- [x] max taker_vol should be udpated once the swap is ongoing (2/3) (your max_slider_volume value should change at this step)

### DIFF
--- a/src/core/atomicdex/models/qt.orders.model.cpp
+++ b/src/core/atomicdex/models/qt.orders.model.cpp
@@ -15,8 +15,8 @@
  ******************************************************************************/
 
 //! Project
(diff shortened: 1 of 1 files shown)


smk762:
Does change slider value mid swap :+1: 
Max slider not updating when funds send out.
image

Milerius:
> Does change slider value mid swap 👍
> Max slider not updating when funds send out.
> image

Yes it's a bit hard for me to find a solution, just using swap event for that specific case

PR #979, opened 2021-06-12
enhancement(switch_mode_ui): add hover effect in switch mode header

ssakone:


tonymorony (approved):
the animation itself lgtm, however app lag now on button switch (what cause animation looks a bit laggy also)

would be great to make tabs switching smooth during trading view polishing @ssakone

PR #988, opened 2021-06-15
Best orders simplified implementation in Simple View

ssakone:


Milerius:
I merged it but still few things to do in the simple view pr: colors

PR #1000, opened 2021-06-22
feat(dex): rework buy side

Milerius:
to test on pro view/buy side

- [x] try to create your own order
- [x] try to create your own order with a min_vol
- [x] try to swallow an order from the orderbook
- [x] try to match an order from the orderbook without swallowing it
- [x] Handle min vol for buy side


smk762:
Created Buy Orders with min vol that reverted to maker orders after failing to match (with and without min_volume).

Swallowed buy order from clicking on orderbook :heavy_check_mark: 

Created buy order manually with same values as orderbook, also swallowed. :heavy_check_mark: 

`try to match an order from the orderbook without swallowing it` no luck with this one yet.

Should we perhaps include the min_volume value on the confirm panel and order detail panel?
e.g.
image

smk762 (dismissed):

cipig:
image

the following case is not matching:
1. go to BIDR/KMD
2. select cheapest offer
3. change volume to 1000 BIDR
4. click "start swap"

reason: the order has `"base_min_volume": "54827.4816590526",` set
if i select an order from orderbook, can it be made so i can't change volume and set it below `base_min_volume` of the selected order?

Milerius:
@cipig @smk762 can you test again to check min_vol handle ?

cipig:
image
works good now, thanks

cipig (approved):

PR #1006, opened 2021-06-22
fix(addressbook): Hot fixes

SylEze:
- [ ] Use [QtQuick 2.15 TableView](https://doc.qt.io/qt-5/qml-qtquick-tableview.html) instead of QtQuick.Controls 1.4 TableView (for Qt 6 migration)
- [ ] Add support for BEP-20 addresses
- [x] No more text overflow handled
- [ ] Use validate/convert address from wallet page directly in addressbook frontend
- [ ] Add a fix button for address that need a conversion (same as wallet approach)

### DIFF
--- a/atomic_defi_design/qml/Wallet/AddressBookAddContactAddressModal.qml
+++ b/atomic_defi_design/qml/Wallet/AddressBookAddContactAddressModal.qml
@@ -33,10 +33,12 @@ BasicModal {
         title: isEdition ? qsTr("Edit address entry") : qsTr("Create a new address")
 
         // Wallet Type Selector
-        DefaultButton {
-            implicitWidth: parent.width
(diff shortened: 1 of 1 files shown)


smk762:
Still not seeing BEP-20 as an option in addressbook
image

Milerius:
> Still not seeing BEP-20 as an option in addressbook
> image

@smk762 syl do not use the same approach, the task list he have is the task left before you can start a review

smk762:
There is no overflow in coin type selection, but all languages return English for that selector now.
image

Milerius:
> There is no overflow in coin type selection, but all languages return English for that selector now.
> image

Its normal he changed the text so need you translation, unrelated to the pr

PR #1037, opened 2021-06-25
fix(trading): Min selector confusing

### LINKED ISSUE #1012: [BUG]: Min volume selector is confusing
At the moment it's not clear what dots do and there is no visual representation of min trading volume (as well as an explanation of what min volume is/using for).

I think we can make it an optional input field (with activation way similar to custom dpow protection rules) with an indicator in percents from 0%  to 100% from order amount

SylEze:
Closes #1012

### DIFF
--- a/atomic_defi_design/qml/Exchange/Trade/TradeBox/OrderForm.qml
+++ b/atomic_defi_design/qml/Exchange/Trade/TradeBox/OrderForm.qml
@@ -169,7 +169,7 @@ FloatingBackground {
 
                 first.value: parseFloat(API.app.trading_pg.min_trade_vol)
 
-                firstDisabled: true
+                firstDisabled: !_useCustomMinTradeAmountCheckbox.checked
                 defaultFirstValue: parseFloat(API.app.trading_pg.min_trade_vol)
                 
                 firstTooltip.text: qsTr("Minimum volume: %1").arg(General.formatDouble(first.value, General.getRecommendedPrecision(second.value)))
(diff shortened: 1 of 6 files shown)


smk762:
If sliding up the minimum, then unticking checkbox, minimum stays but moveable dot hidden.
image

Expected: If not using custom minimum, line and value should revert to 0.

ssakone:
@smk762 can you test again after CI?

atomic_defi_design/qml/Exchange/Trade/TradeBox/OrderForm.qml
```diff
@@ -230,6 +233,18 @@ FloatingBackground {
...
+                        _volumeRange.oldFirstValue =  _volumeRange.defaultFirstValue
+                    } 
+                    else { 
+                        _volumeRange.defaultFirstValue = API.app.trading_pg.orderbook.current_min_taker_vol
+                        _volumeRange.first.value = API.app.trading_pg.orderbook.current_min_taker_vol
+                    }
+                }
+                onCheckedChanged: {
+                    
+                }
```
Milerius, line 247: What the purpose of this code ? Why do you push code that literally does nothing ?

smk762 (dismissed):
> If sliding up the minimum, then unticking checkbox, minimum stays but moveable dot hidden.
> image
> 
> Expected: If not using custom minimum, line and value should revert to 0.

Confirm working as expected now

atomic_defi_design/qml/Exchange/Trade/TradeBox/OrderForm.qml
ssakone: sorry.

PR #1039, opened 2021-06-25
fix(max_taker_vol): fees show behaviour

Milerius:
- [x] fees always shows on confirmation screen
- [x] If you are on simple view fees should always be visible after selecting an order

Please check logs if it's not the case, you should have enough informations


tonymorony (approved):
it looks like the issue resolved now, thanks

smk762 (approved):

PR #1042, opened 2021-06-25
fix(trading): Default view is simple after a reset cfg

### LINKED ISSUE #1034: [BUG]: Simple view should be the default view after a reset
**Describe the bug**

In the backend when i create cfg.ini i do `create_settings_functor("DefaultTradingMode", TradingMode::Simple);`

This variable should be used no matter what in the front end to determine default view (should be Simple by default)

SylEze:
Closes #1034


smk762 (approved):
Working as expected here :+1:

PR #1043, opened 2021-06-25
fix(trading): Simple View -> Selected coin was disabled in coin selection list

### LINKED ISSUE #995: [BUG] Simple Trading View: selected coin disabled in coins selection view dropdown menu
If user has selected coin to sell, and then clicks to open the selection menu again, the selected coin within the list is not active / selectable if clicked on. So if user goes to select different coin and then changes their mind to keep it the same as it was, they need to click the "X" in the top right.

https://user-images.githubusercontent.com/35845239/122749188-9d9f9580-d2bf-11eb-963d-efcd324c5227.mp4

Expected behaviour would be user can click on the coin already selected in the list, and return to the simple form without clicking "x".

_Originally posted by @smk762 in https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/889#issuecomment-864929928_

SylEze:
Close #995


smk762 (approved):
Working as expected

PR #1063, opened 2021-07-07
feat(segwit): add segwit support

Milerius:
- update mm2
- add cfg field to determine if segwit exist and if it's on


smk762:
Confirm mm2 version update in settings :heavy_check_mark: 
Segwit toggle present :heavy_check_mark: 

image

Should be `proceed` instead of `procede`.

Message is a bit ambiguous - does it mean to send funds from legacy to segwit? I see it auto-populates the "To address" field with what I assume is my segwit address, but as it is not explicitly stated, I am frightened and start having flashbacks about the time I got pizza delivered to my old house after moving out. 

Suggest something like - 
```
Do you want to send your {this_addr_type} funds to {next_addr_type} wallet first?
```

Though that would imply it can only go to your related address, and the input field is not read-only. So maybe make that field read only in this case?

Alternatively, make message -
`Do you want to send your {this_addr_type} funds first?` 

with options (instead of yes/no)
`To {new_addr_type} address | To other address | No`  

Tbh, the popup is a bit annoying if it happens all the time, I'd almost prefer it want there at all.

Until balance updates after toggling, suggest some "in progress / loading" animation like spinning dots.

smk762:
I'm also seeing notification for balance change when switching between legacy and segwit.
image

Milerius:
@ssakone permanent wrong values send to backend side please test everycase, check carefully the logs, write mock unit test if you have any doubt of what you are trying to do, or ask question

smk762:
image
Instead of addresses, these variables should be wallet type (e.g. Legacy || Segwit)

smk762:
> image
> Instead of addresses, these variables should be wallet type (e.g. Legacy || Segwit)

Fixed :heavy_check_mark:

smk762 (approved):

Milerius:
@smk762 Let's wait the merge of the pr from atomicdex api before we merge this one, i use a specific branch of mm2 from @shamardy atm

Milerius:
@ssakone can you use your new dialog here ?  instead of the one from qaterial

src/core/atomicdex/pages/qt.wallet.page.cpp
```diff
@@ -852,4 +866,106 @@ namespace atomic_dex
...
+            const auto ticker   = get_current_ticker().toStdString();
+            const auto coin_cfg = mm2_system.get_coin_info(ticker);
+            if (coin_cfg.segwit)
+            {
+                nlohmann::json address_format = nlohmann::json::object();
+                address_format                = {{"format", "segwit"}};
+                if (!checked)
+                {
+                    //! We go from segwit to legacy
+                    if (coin_cfg.ticker != "BCH")
```
tonymorony, line 886: what about LTC, BTC, <othercoins with segwit support>?

PR #1070, opened 2021-07-08
enhancement(hide_balance): fix balance display in header

ssakone:


smk762 (approved):
Working as expected on all OS

PR #1075, opened 2021-07-11
feat(window_icon): add window_icon in prerequisites

Milerius:


smk762:
Doesn't appear to be working, though I cant find error in logs. 

Perhaps needs to be after this line - https://github.com/KomodoPlatform/atomicDEX-Desktop/blob/e228b348207db2c9009977a966b0b594968852c8/src/app/main.prerequisites.hpp#L440

for the qrc: path prefix to work

smk762 (approved):

PR #1089, opened 2021-07-13
feat(orders_model): add recover_funds in this model in an asynchronous way

Milerius:
…s way


Milerius:
(only front-end part left)

smk762:
Still freezes until crash, pls ping me when front end is done @SylEze @Milerius

smk762 (approved):
no longer blocking app, and no crash. err msg returned after some time. mm2 functioning ok afterwards to make a swap.

PR #1091, opened 2021-07-13
feat(portfolio): Show number of active coins in portfolio

### LINKED ISSUE #821: [FEATURE REQUEST]: Add warning when no balances and "Show only coins with Balance" toggle on
On new wallets without any balances, if the show balance toggle is on, the table at the bottom is empty
image

This makes sense, but can lead to a brief "where are my coins!" moment.  In this state, the "Add new asset" button is also not present.
image

Suggest some warning text appear next to toggle if it is on and no balances. e.g. 
image

Alternatively (or additionally) it might be worth specifying how many activated coins have balance when this toggle is on.
image

SylEze:
Closes #821

### DIFF
--- a/atomic_defi_design/assets/languages/atomic_defi_en.ts
+++ b/atomic_defi_design/assets/languages/atomic_defi_en.ts
@@ -1162,6 +1162,50 @@
     </message>
 </context>
 <context>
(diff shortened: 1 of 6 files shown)


smk762 (approved):
Looks good, thanks

PR #1105, opened 2021-07-16
feat(portfolio): Show asset type in amounts chart

### LINKED ISSUE #1076: [BUG]: Duplication in portfolio percentages
**Describe the bug**
When you have balance in two varieties of the same coin (e.g. BTC and BTC-BEP20), these are both shown as BTC, but in separate lines

image

**To Reproduce**
1. Have BTC and BTC-BEP20 balance (or other coin with more than one variety)
2. Login and go to dashboard
4. See duplication

**Expected behavior**
Two options: 
- combine both balances into one row
- or, include "BEP20" etc in label where appropriate.

SylEze:
Closes #1076

### DIFF
--- a/atomic_defi_design/assets/languages/atomic_defi_en.ts
+++ b/atomic_defi_design/assets/languages/atomic_defi_en.ts
@@ -1384,33 +1384,71 @@
 <context>
     <name>FirstLaunch</name>
     <message>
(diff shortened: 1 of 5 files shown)


smk762 (approved):
Confirm working as expected

PR #1124, opened 2021-07-19
enhancement(wallet_set_fix): Change recover textarea to textField, remove wallet deletion in settings

ssakone:


smk762 (commented):
Confirm "Delete wallet" removed from settings.
Confirm input for wallet recover is not single line with hidden input visibility. 
Is this possible with text area? Do we want to be able to see all words input without scrolling around?

ssakone:
> Confirm "Delete wallet" removed from settings.
> Confirm input for wallet recover is not single line with hidden input visibility.
> Is this possible with text area? Do we want to be able to see all words input without scrolling around?

qml TextArea & TextEdit not support this, maybe if i create new component based TextEdit (not sure, i need to test)

ssakone:
> Confirm "Delete wallet" removed from settings.
> Confirm input for wallet recover is not single line with hidden input visibility.
> Is this possible with text area? Do we want to be able to see all words input without scrolling around?

otherwise we can just add two field, when text mode == hidden, we display textArea, when hidden we display just hidden field

smk762 (approved):

tonymorony (approved):

PR #1126, opened 2021-07-19
feat(gecko): add total_volume parsing

Milerius:


tonymorony (approved):

tonymorony (changes requested):
would be great to filter smartchains

tonymorony (approved):

PR #1132, opened 2021-07-22
feat(recover_funds): more precise recover funds, and use good http st…

Milerius:
…atus


smk762:
image

I get same as before. Should I try and simulate a new failed swap?

tonymorony:
jfyi: PR got conflict with translation now also
`atomic_defi_design/assets/languages/atomic_defi_fr.ts`

tonymorony:
@smk762 for my old swaps it printing valid message (not very user-friendly tho):
image

Milerius:
> image
> 
> I get same as before. Should I try and simulate a new failed swap?

Yes it's expected for this one

Milerius:
> @smk762 for my old swaps it printing valid message (not very user-friendly tho):
> image

In the future with mm2 2.0 rpc call error message will be improved because it's will be separated into different field and easier to translate for example.

tonymorony:
`In the future with mm2 2.0 rpc call error message will be improved because it's will be separated into different field and easier to translate for example.`

I meant not the error message itself but error decoration, print plain JSON is not a good practice in GUI. Unlikely you'll see in production-ready desktop application error in plain text. But yeah, it's unrelated to this PR

smk762 (approved):

PR #1139, opened 2021-07-23
enhancement(seed_confirmation): Add Dialog input before Allow custom seed

ssakone:


smk762 (approved):
Looks good here, thanks.

PR #1144, opened 2021-07-25
Fix various bug

ssakone:
Check List

- [ ] Proview filtering fixed
- [ ] Checkbox not overflow in order place form.


smk762 (approved):
Looks ok here, thanks.

PR #1147, opened 2021-07-26
feat(security): seed length was limited by the component length field limit

ssakone:


Milerius:
tested thanks for the prompt fix

PR #1152, opened 2021-07-29
fix(portfolio_fix): Move Portfolio Pie List item from Column Repeater to ListView

ssakone:


ssakone:
also reformated in my theming PR

ssakone:
@smk762  check it again, if the bar size change when we disable & enable, if its work maybe its back-end issue.

smk762 (approved):
Works as expected here, thanks.

PR #1156, opened 2021-08-02
feat(netid): add possibiltiy to change netid for smartdex

Milerius:


src/app/main.prerequisites.hpp
```diff
@@ -333,6 +333,10 @@ handle_settings(QSettings& settings)
     create_settings_functor("MaximumNbCoinsEnabled", 50);
     create_settings_functor("DefaultTradingMode", TradingMode::Simple);
     create_settings_functor("FontMode", QQuickWindow::TextRenderType::QtTextRendering);
+    if (QString{DEX_NAME} == "SmartDEX")
+    {
+        create_settings_functor("NetID", 7777);
```
tonymorony, line 338: in the case of netid switch you'll also need to switch "seednodes" param since seednodes only for 7777 netid are in API code

Milerius: So what do we prefer between setting extra seednodes or setting `i_am_a_seed` to `true` ?

In the first case could you please give a list of valid seednode on another network to test ?

Milerius:
Tested manually by merging on smartdex branch locally - works correctly, so what we do for the seednodes ?

tonymorony:
Not intended to be merged into d

PR #1161, opened 2021-08-04
feat(theming): override theming at each run

Milerius:


smk762:
> _No description provided._

So... I'm testing that this overwrites the themes json on each launch?

smk762:
image
Saw this when backing up my themes. Folder empty except "old" subfolder and zip archive of them I attached in another PR.

We should reduce this list to only display .json files - and maybe add validation / error message if attemping to load something invalid.

```
[19:51:57] [warning] [main.prerequisites.hpp:92] [422518]: QString::arg: Argument missing: END APPLY , old
[19:51:57] [debug] [main.prerequisites.hpp:86] [422518]: END APPLY 
[19:52:03] [warning] [main.prerequisites.hpp:92] [422518]: QString::arg: Argument missing: END APPLY , smk_light.zip
[19:52:03] [debug] [main.prerequisites.hpp:86] [422518]: END APPLY 
```

smk762:
If I delete/move the theme json files out of their normal folder, they are not recreated on app launch.
If exiting app with an alternative theme, it is not reset to default on next launch.

Still not sure what I'm testing here

Milerius:
> If I delete/move the theme json files out of their normal folder, they are not recreated on app launch.
> If exiting app with an alternative theme, it is not reset to default on next launch.
> 
> Still not sure what I'm testing here

you do not have to delete the content of the theme folder, it's not intended - it's just the json file that are overriden not the settings

tonymorony:
> > If I delete/move the theme json files out of their normal folder, they are not recreated on app launch.
> > If exiting app with an alternative theme, it is not reset to default on next launch.
> > Still not sure what I'm testing here
> 
> you do not have to delete the content of the theme folder, it's not intended - it's just the json file that are overriden not the settings

it would be very helpful if you can create a testing checklist in the PR description on opening

PR #1162, opened 2021-08-04
enhancement(various_fix2): Fix notification persistence & Make order ID copyable.

ssakone:


smk762:
notifications fixed :heavy_check_mark: 
Copy works, but the only indication uuid is active is cursor change when you hover over it - might be better to use a copy symbol like how it is on wallet retrieve to copy address (and highlighting uuid when clicked)

Also, the "uuid" copied message is also popping up under the swaps detail panel so it is mostly obscured
image

smk762:
Copy icon working as expected :heavy_check_mark: 
Pop up message still buried :crossed_flags: 
image

ssakone:
@smk762  need review

smk762 (approved):
Working as expected

tonymorony (approved):

PR #1164, opened 2021-08-05
Update atomicdex-desktop-ci.yml

smk762:


smk762:
@tonymorony @Milerius 3rd run completed - qt mirror is ok, but had a couple of unrelated fails in test phase on OSX. I believe this was an temporary issue with github but will continue to monitor.

If we can approve / merge this, should at least give us artefacts to test the themeing PR etc.

PR #1178, opened 2021-08-10
add NVC-BEP20

cipig:


smk762 (approved):
Sighted in app after assets reset

Issue #1181, opened 2021-08-10
[BUG]: Dancing funds recovery button


tonymorony:
https://user-images.githubusercontent.com/24797699/128879225-b866b772-c53f-4839-8cd3-d98be20e5119.mov

Issue #1180, opened 2021-08-10
[BUG]: Pro view filter persists in Simple view


smk762:
**Describe the bug**
Applying a filter in pro view (e.g. KMD/DASH), applies the same filter in simple view, but does not update the filter summary text at the top of the table (reverts to ALL/ALL).

**To Reproduce**

https://user-images.githubusercontent.com/35845239/128858294-6f06505b-07a3-4d60-be35-ca7cd88e2789.mp4

**Expected behavior**
Filter should reset to default when switching from pro view to simple view.

**Additional context**
The opposite case (switching from simple to pro view) the filter is reset to default as expected.

PR #1183, opened 2021-08-11
Fix:  Recover Fund  button, Simple View filtering

ssakone:
# Check List

- [ ]  #1181 
- [x] #1180


smk762:
I think #1180 is already fixed in https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/1182

I get same button resize as in issue when clicking on recover funds.

ssakone:
I will remove it.

PR #1184, opened 2021-08-11
update mm2 for smoke testing

tonymorony:
https://github.com/KomodoPlatform/atomicDEX-API/pull/1004#issuecomment-875517736


smk762:
Launch and simple swap done on all platforms, lmk if more smoke req.

PR #1215, opened 2021-08-26
feat(lint): start linting the code

Milerius:
- try a swap rick/morty
- try standard trading feature


smk762 (approved):
Does trade without problem

PR #1222, opened 2021-08-30
feat(segwit): re-enable segwit

Milerius:
steps: 

- Reset assets configuration
- Switch between LTC segwit / legacy 

give any feedbacks if any


smk762:
image

Should be "sent" instead of "send". 

Replace text with "Your transaction has been sent, it may take a little while to arrive."

Milerius:
regarding UI/UX does it use the last modal @ssakone ?

artemii235 (changes requested):
Having the `Segwit` toggle in the specific coin dashboard is quite an unexpected way of implementation. I assumed that there would be a `Select Address Format` dropdown on the initial coin activation screen. 

Alternatively, as discussed with @tonymorony and @yurii-khi the address format selection can be done on the wallet creation/import screen without the possibility to change it. It's done this way in the Electrum  Bitcoin wallet. 

Also, we can have more address formats supported in the future - the toggle can't handle this.

image

Milerius:
> Alternatively, as discussed with @tonymorony and @yurii-khi the address format selection can be done on the wallet creation/import screen without the possibility to change it. It's done this way in the Electrum Bitcoin wallet.

It's will actually simplify the implementation a lot for us.

thanks for the feedback !

artemii235:
> It's will actually simplify the implementation a lot for us.

@Milerius Please note that users can have LTC on the legacy address and BTC on the Segwit address. So we can't set the default address format for all the coins at once.

Milerius:
> > It's will actually simplify the implementation a lot for us.
> 
> @Milerius Please note that users can have LTC on the legacy address and BTC on the Segwit address. So we can't set the default address format for all the coins at once.

In this case it's much more complicated.

Why not transforming the toggle into the `address format to choose` then and keep the same behaviour as now ? 

For example with a drop down list and let the user choose his favorite format ?

smk762:
Interesting to note - after testing this and toggling LTC to segwit, I exited app and loaded another branch without the toggle displayed. When logging into the same wallet as before, my LTC was still set to segwit. 

I think from a UX perspective it would be nice to have both options enabled / visible / usable within same wallet screen

In comparison with CLI, total balance is shown and user only realise some are legacy and some are segwit when checking addressgroupings or utxos.

If possible and not too tricky to code, showing combined balance on wallet page / dashboard would be nice, generally for the purposes of spending user doesnt care which address funds are sent from.

Clicking the receive button could show both QR codes, tagged by type.

Transaction history might be more of a challenge, may need to combine from two or more queries then sort by date before populating on screen.

Related:
https://github.com/KomodoPlatform/atomicDEX-API/issues/720 (due to this, my mined TKL is not shown until sent to self)
https://github.com/KomodoPlatform/atomicDEX-API/issues/1006

smk762:
> Related:
> [KomodoPlatform/atomicDEX-API#720](https://github.com/KomodoPlatform/atomicDEX-API/issues/720) (due to this, my mined TKL is not shown until sent to self in CLI)

FWIW this also makes tx history look a bit wierd, like I've never received funds but have spent them
image

artemii235:
> In this case it's much more complicated.
Why not transforming the toggle into the address format to choose then and keep the same behaviour as now ?

@Milerius 

Once the address format is chosen, it's very undesirable to allow users to change it so easily:
1. If users click the toggle (or dropdown) occasionally their balance will switch to 0. It can cause a lot of frustration.
1. When a swap is running, the coins involved can't be deactivated. The address format switch is impossible during this time.
1. If users manage to switch address format between MM2 restarts, and they have unfinished swaps, their funds will be sent to another address (not with which the swap started). So we can get a situation when they will have a balance on both Segwit and Legacy addresses, which will also cause frustration.

PR #1225, opened 2021-08-30
fix(trade): Cancel order button was dancing in simple view

### LINKED ISSUE #1220: [BUG]: Simple view orders dancing
**Describe the bug**

https://user-images.githubusercontent.com/21139416/131240426-3342e697-f49b-4c2a-a5e6-de671a4f8c50.mov

SylEze:
* `visible` condition was using `hovered` value, which should not be the case because this is unrelated. Indeed, in that case, cancel order button should be displayed only if the order is cancellable, not if the mouse is hovering the button.

Fixes #1220


smk762 (approved):
Working as expected here

PR #1229, opened 2021-08-30
add couple gecko/paprika ids

cipig:


smk762 (approved):
modified coins sighted with price. WWCN sighted (no price tho)

cipig:
i took the geckoid for WWCN directly from https://www.coingecko.com/en/coins/wrapped-widecoin
no price either because https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/1233 was not added to this branch by the time i made the PR or because the price on coingecko is 12 days old... not sure

smk762:
Paprika also tags it as "inactive" so both may be returning a zero balance because of stale data

PR #1232, opened 2021-08-31
Wallet enabling rework

SylEze:
Will close https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1212 when ready.

It's in a draft stage for the moment.


Milerius:
You can start checking @smk762 

i suggest:

- try with 5-10 coins enabled, logout/login multiple times
- same with 50~coins
- same with 100~coins

smk762:
- batch all parent coins first
- batch the rest in groups of 5 or 10

SylEze:
Detected various frontend issues:

- [ ] Notifications panel UX not good
- [ ] Dashboard "Show only coins with balance" switch also affects the wallet page which should not be the case

Milerius:
I found the reason of the slowing bug when enabling lot of coins, it's due too:

`src/core/atomicdex/models/qt.portfolio.model.cpp:42`

- `m_model_proxy->setDynamicSortFilter(true);`

This property make Qt dynamically recalculate the whole sorting/filtering each time a bunch of coins are added, and this actions can be slow if repeated too often.

Few ideas how to solve that:

- At construction, set this property to false
- When the number of active coins reach the number of enabled coins, re-enable it to true
- If the user require to add too much coins at once (let's say more than 20) then we need to turn off this property to off again, and post enabling, re-enable it + resort.

can you take care of that @SylEze ?

smk762:
Balance from enable works great, seen as soon as added to list.

tonymorony (approved):
that's resolving the problem of indefinitely hang of app startup with a lot of assets 

however, there are two more problems which we need to resolve @SylEze @Milerius :

1) if you start the app with 100-200 assets app became unresponsive/hangs for 30 seconds or so
2) when there are a lot of assets there is a loading spinner on tabs switching to the dashboard

PR #1233, opened 2021-08-31
feat(coingecko): handle fiat values for incomplete coins answer

Milerius:


smk762:
context?

Milerius:
> context?

The KOIN coin didn't have fiat value before

smk762 (approved):
KOIN fiat value sighted

PR #1242, opened 2021-09-01
Auto open place order form when order selected

ssakone:


smk762:
Returns buy form if it was closed, but stays hidden if it was hidden

smk762:
Should work ok now.
Form should reappear if:
 - closed order form, then click on order in orderbook
 - hidden order form, then click on order in orderbook
 - closed order form, then click on order in bestorders
 - hidden order form, then click on order in bestorders

tonymorony (approved):
lgtm, thanks!

PR #1250, opened 2021-09-04
add Axie Infinity (AXS)

cipig:


smk762 (approved):
all coins in menu, charts display verified and no problems with activation or price

PR #1254, opened 2021-09-06
enhancements(naming): simple -> SmartSwap, pro -> SmartDEX

Milerius:


smk762:
Got some button overlap
image

smk762:
Confirm lite mode as default after assets reset

Milerius:
Hey @SylEze can you give me some advice on this one ? I've tried to change the spacing property, but that make things worst, i see there is this `Item` above which is likely the rectangle around the label text, but somehow i don't manage to make it follow the same width as the text size

PR #1256, opened 2021-09-06
fix(crash): fix a crash when quitting the application on the trading …

Milerius:
…page


smk762 (approved):
confirm clean exit from dex pro screen

PR #1261, opened 2021-09-07
0.5.1

ca333:


Milerius (changes requested):
After testing this morning i've 2 suggestion:

bump mm2.1 with the memory fix commit: https://github.com/KomodoPlatform/atomicDEX-API/issues/941#issuecomment-914924876 and update the mm2 release here

remove band oracle usage, they produce too much price difference from the market, doing my routine test this morning i can clearly say that the coingecko result are more reliable.

Ravencoin was show as 0.14$ when it's was only 0.119$ in reality. (coingecko, binance and so on was showing the good price, but the oracle was clearly with non-existent price.)

tonymorony:
> After testing this morning i've 2 suggestion:
> 
> bump mm2.1 with the memory fix commit: [KomodoPlatform/atomicDEX-API#941 (comment)](https://github.com/KomodoPlatform/atomicDEX-API/issues/941#issuecomment-914924876) and update the mm2 release here
> 
> remove band oracle usage, they produce too much price difference from the market, doing my routine test this morning i can clearly say that the coingecko result are more reliable.
> 
> Ravencoin was show as 0.14$ when it's was only 0.119$ in reality. (coingecko, binance and so on was showing the good price, but the oracle was clearly with non-existent price.)

have to re-test 3 fixed issues on the API side:

https://github.com/KomodoPlatform/atomicDEX-API/issues/1055
https://github.com/KomodoPlatform/atomicDEX-API/issues/941
https://github.com/KomodoPlatform/atomicDEX-API/issues/1024

then we'll be able to finalize the API version for this patch

smk762 (approved):
Launches on all OS, simple smoke tests passing

tonymorony (approved):

PR #1293, opened 2021-09-15
Restore swap status text in swap/order details modal

smk762:
Resolves https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1258

Extra whitespace was due to missing elements (swap status and loading spinner) which failed to draw because a function was missing.

image


tonymorony:
thanks! it would be great to make it looks more like on figma 

Screenshot 2021-09-16 at 12 57 35
Screenshot 2021-09-16 at 12 57 20

image

tonymorony (approved):

PR #1294, opened 2021-09-15
Add label when no result in simple view history and orders

### LINKED ISSUE #1284: [FEATURE REQUEST]: Add "No results found" text in case of empty filter result
In simple and pro trading view filters when you do a search and no results are found the user doesn’t get any response back it just leaves the search in blank. “No results found “ should be added when no results are generated.

image

SylEze:
Closes #1284


smk762:
I see this working on simple view filters, but not on Pro view filters.

SylEze:
> I see this working on simple view filters, but not on Pro view filters.

Added in latest commit

smk762 (approved):

PR #1305, opened 2021-09-17
add number suffixes for seed confirmation

smk762:
- Resolves https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/629
- functionized reset action

### DIFF
--- a/atomic_defi_design/qml/Screens/NewUser.qml
+++ b/atomic_defi_design/qml/Screens/NewUser.qml
@@ -34,6 +34,19 @@ SetupPage {
         return current_mnemonic.split(" ")
     }
 
+    function getNumberTh(num) {
+        switch (num) {
+            case 1:
+                return 'st'
+            case 2:
+                return 'nd'
+            case 3:
+                return 'rd'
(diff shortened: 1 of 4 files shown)


atomic_defi_design/qml/Screens/NewUser.qml
```diff
@@ -34,6 +34,19 @@ SetupPage {
         return current_mnemonic.split(" ")
     }
 
+    function getNumberTh(num) {
+        switch (num) {
+            case 1:
+                return 'st'
```
SylEze, line 40: I think you need to use `qsTr()` on those strings since thoses nb suffixes are not the same across diferent languages.

SylEze (changes requested):

atomic_defi_design/qml/Screens/NewUser.qml
smk762: fixed

smk762:
Also resolves https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/400

PR #1324, opened 2021-09-20
add FJCB Fujicoin

fujicoin:
It was found that using the current FJC-BEP20 token at a 500: 1 exchange rate would cause inconvenience in many cases.
Therefore, I decided to use FJC-BEP20 for a 1: 1 rate and issue a new token FJCB for 500: 1.
ContractID: `0xda73AE86D38D9b5D22C64722320cCC071cB79ba3`

Currently there is a problem with the next line.
```
     "coinpaprika_id": "fjc-fujicoin",
     "coingecko_id": "fujicoin",
```
This is corrected by acquiring a new ID after creating a transaction record with ADEX or Pancake Swap.


fujicoin:
I fixed coinpaprika_id and coingecko_id

tonymorony (approved):

PR #1364, opened 2021-10-10
change price endpoint

cipig:


tonymorony:
do we need this PR opened?

PR #1389, opened 2021-10-31
feat(matic): support gas matic station decimals

Milerius:


cipig (approved):
works good, thanks for the fix

PR #1411, opened 2021-11-08
Sidebar cursor not refreshing when switching page outside of the sidebar

SylEze:


smk762:
Still seems to not shrink for me here on linux (local build, also tried assets reset with no luck)

PR #1429, opened 2021-11-15
update QTUM icon

cipig:


smk762:
Do we need to update tQTUM also?
image

Image canvas does not have a square shape, so there is some squeezing effect on some pages.
image

Icon appears smaller than others in some places. Could be related to above, we should "clip to content" to remove and extra whitespace at top/bottom/sides then square it off.
image

Milerius:
Let's wait for a better image quality from the team

PR #1437, opened 2021-11-18
apply gradient to disabled button in simple dex swap

smk762:
Before
image

After
image


smk762:
Before
image

After
image

PR #1449, opened 2021-11-23
add TRYB-PLG20

cipig:


smk762 (approved):
All 3 new coins sighted and enabled without issue

Issue #1448, opened 2021-11-23
[FEATURE REQUEST]: Ensure that RPC password generating by app matching new mandatory criteria 


tonymorony:
e.g. unlikely we checking now for 1 and 7

"
We are going to merge the PR this week (https://github.com/KomodoPlatform/atomicDEX-API/pull/1144) that applies additional rules to the RPC password set in the config as follows:

1. It can't contain the word "password".
2. Its length should be between 8 and 32 characters.
3. It should contain at least 1 digit.
4. It should contain at least 1 lowercase character.
5. It should contain at least 1 uppercase character.
6. It should contain at least 1 special character.
7. It can't contain the same character 3 or more times in a row.

Please change your RPC password accordingly on the API update from mm2.1 or dev branches after this is merged. This change will also be released on Github this week. If you prefer to turn off these restrictions, please add "allow_weak_password":true in the MM2.json. 
"

PR #1450, opened 2021-11-24
feat(password_policy): start transition to new password policy

Milerius:
Checklist:

- [ ] Check that generated rpc password respect all criteria from #1448 .


smk762 (approved):
Version sighted in app: `2.1.4829_mm2.1_762f40258_Linux_Release`
Logged in 12 times, in each case the rpc_password generated was acceptable.

tonymorony (approved):

PR #1452, opened 2021-11-27
add couple price ids

cipig:


smk762 (approved):
Loads ok and coins activate without issue.

PR #1455, opened 2021-11-29
feat(settings): check for url local or scheme before processing token

Milerius:


tonymorony:
@Milerius is it still wip? we'll freeze RC code on monday

Milerius:
> @Milerius is it still wip? we'll freeze RC code on monday

Yes unfortunately because it requires front end changes to handle error earlier

tonymorony:
looks stalled

PR #1469, opened 2021-12-03
Header & notification redesign

ssakone:


tonymorony:
lgtm, thanks!

the only thing I noticed is that there is no hover effect on notification in light theme

tonymorony (approved):

SylEze (approved):

PR #1471, opened 2021-12-03
ZET chain swap

cipig:


smk762 (approved):
Both coins enable without problem.
urls in coins.json are ok.
KMD/ZET swap done without problem. TX displays in history.
Unable to source ZET-OLD, but looks ok.

PR #1481, opened 2021-12-06
feat(version): bump to 0.5.4

Milerius:


smk762 (approved):
Launch and coin activation ok

PR #1490, opened 2021-12-09
feat(windows): mt the manifest

Milerius:


smk762 (approved):
installer is now functioning like all other binaries.

PR #1515, opened 2021-12-16
Change ticker selector hover color in simple trading view

SylEze:


tonymorony:
up and bottom selector hover color is different now in both dark and light theme, it looks like from coin selector hover is not ok and it'll be fine if the same color as for order selection hover will be used

image
image

SylEze:
> 

Sorry fixed that, colors were hardcoded

tonymorony (approved):

smk762 (approved):

PR #1520, opened 2021-12-16
add chain_id for custom tokens

smk762:
closes: https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1361

### DIFF
--- a/src/core/atomicdex/pages/qt.settings.page.cpp
+++ b/src/core/atomicdex/pages/qt.settings.page.cpp
@@ -431,14 +431,6 @@ namespace atomic_dex
             out["mm2_cfg"]      = nlohmann::json::object();
             out["adex_cfg"]     = nlohmann::json::object();
             const auto& mm2     = this->m_system_manager.get_system<mm2_service>();
-            if (parent_chain == "BNB")
-            {
-                out["mm2_cfg"]["chain_id"] = 56;
-            }
-            else if (parent_chain == "ETH")
-            {
-                out["mm2_cfg"]["chain_id"] = 1;
-            }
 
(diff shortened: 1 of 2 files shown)


src/core/atomicdex/pages/qt.settings.page.cpp
```diff
@@ -431,12 +431,22 @@ namespace atomic_dex
...
             const auto& mm2     = this->m_system_manager.get_system<mm2_service>();
+            if (parent_chain == "BNB")
+            {
+                out["mm2_cfg"]["chain_id"] = 56;
+            }
+            else if (parent_chain == "ETH")
+            {
+                out["mm2_cfg"]["chain_id"] = 1;
+            }
+
```
Milerius, line 442: I'm not a fan of hardcoding it - raw_parent_cfg is made for that purpose, also it's better to be in the inner if line 444.

Milerius (changes requested):

src/core/atomicdex/pages/qt.settings.page.cpp
smk762: `raw_parent_cfg` returns `{"avg_blocktime":0.05,"coin":"BNB","fname":"Binance Coin","mm2":1,"name":"binancesmartchain","protocol":{"type":"ETH"},"required_confirmations":3,"rpcport":80}`

`chain_id` not within.

cipig (approved):
works fine, 0.5.4/configs/coins.json contains chain chain_id: 56 after adding a custom BEP20 token

Milerius (approved):

PR #1522, opened 2021-12-17
use DefaultImage

smk762:
`DefaultImage` extends image with
```
    mipmap: true
    fillMode: Image.PreserveAspectFit
```
So does `DexImage`, though it was only in use in one place. I've updated all references of `Image` and `DexImage` to use `DefaultImage`

closes: https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1351


cipig (approved):
thanks for the fix, now all icons look the same

SylEze (approved):

tonymorony (approved):

PR #1543, opened 2022-01-03
fix smartchain vs komodochain

smk762:
Prior to this change, KMD / RICK / MORTY etc had error for "no QTUM balance" (incorrect parent chain)

### DIFF
--- a/atomic_defi_design/Dex/Constants/General.qml
+++ b/atomic_defi_design/Dex/Constants/General.qml
@@ -357,6 +357,7 @@ QtObject {
         if(type === "ERC-20") return "ETH"
         else if(type === "QRC-20") return "QTUM"
         else if(type === "Smart Chain") return "KMD"
+        else if(type === "Komodo Chain") return "KMD"
         return "?"
     }
 
--- a/atomic_defi_design/Dex/Constants/General.qml
+++ b/atomic_defi_design/Dex/Constants/General.qml
@@ -356,8 +356,7 @@ QtObject {
     function getParentCoin(type) {
         if(type === "ERC-20") return "ETH"
         else if(type === "QRC-20") return "QTUM"
-        else if(type === "Smart Chain") return "KMD"
-        else if(type === "Komodo Chain") return "KMD"
+        else if(type === "Smart Chain" || type === "Komodo Chain") return "KMD"
         return "?"
     }
 
--- a/assets/config/0.5.4-coins.json
+++ b/assets/config/0.5.4-coins.json
@@ -1187,7 +1187,7 @@
   "BET": {
     "coin": "BET",
     "name": "BET",
-    "type": "Smart Chain",
+    "type": "Komodo Chain",
     "coingecko_id": "test-coin",
     "coinpaprika_id": "test-coin",
     "electrum": [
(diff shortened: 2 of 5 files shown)


atomic_defi_design/Dex/Constants/General.qml
```diff
@@ -357,6 +357,7 @@ QtObject {
         if(type === "ERC-20") return "ETH"
         else if(type === "QRC-20") return "QTUM"
         else if(type === "Smart Chain") return "KMD"
```
tonymorony, line 359: shouldn't this check work?
it seems doing the same and initial wording matching this check

smk762: https://github.com/KomodoPlatform/atomicDEX-Desktop/blob/smartdex/src/core/atomicdex/config/coins.cfg.cpp#L102 
vs 
https://github.com/KomodoPlatform/atomicDEX-Desktop/blob/dev/src/core/atomicdex/config/coins.cfg.cpp#L102

Problem is that in `smartdex` branch, there is no match for `Smart chains` in coins.cfg.cpp

A release or two ago it was requested to tag `Smart Chains` as `Komodo Chains` in Smartdex.

In build from https://github.com/KomodoPlatform/atomicDEX-Desktop/actions/runs/1647247078 it was a mix of `Smart chain` and `Komodo chain` and I was seeing errors like this:

image

smk762:
Context: here is where we first did this change in 0.5.1
https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/1257/files

atomic_defi_design/Dex/Constants/General.qml
Milerius: > https://github.com/KomodoPlatform/atomicDEX-Desktop/blob/smartdex/src/core/atomicdex/config/coins.cfg.cpp#L102 vs https://github.com/KomodoPlatform/atomicDEX-Desktop/blob/dev/src/core/atomicdex/config/coins.cfg.cpp#L102
> 
> Problem is that in `smartdex` branch, there is no match for `Smart chains` in coins.cfg.cpp
> 
> A release or two ago it was requested to tag `Smart Chains` as `Komodo Chains` in Smartdex.
> 
> In build from https://github.com/KomodoPlatform/atomicDEX-Desktop/actions/runs/1647247078 it was a mix of `Smart chain` and `Komodo chain` and I was seeing errors like this:
> 
> image

What about:

```
else if (cfg.type == "Komodo Chain" || cfg.type == "Smart Chain")
        {
            cfg.type = "Smart Chain";
            cfg.coin_type = CoinType::SmartChain;
        }
```

in <https://github.com/KomodoPlatform/atomicDEX-Desktop/blob/dev/src/core/atomicdex/config/coins.cfg.cpp#L102>

this way any other change will not be necessary

smk762: I'll do this and see if we can get it into a state where the only change needed is in coins json file (in future we should be able to do this in the background perhaps?)

Milerius: > I'll do this and see if we can get it into a state where the only change needed is in coins json file (in future we should be able to do this in the background perhaps?)

Yes we could do it 100% in background but there's is probably an extra tweak needed in frontend to display Komodo Chain instead of Smart Chain

tonymorony (approved):

PR #1574, opened 2022-01-28
added Doichain

inspiraluna:


atomic_defi_design/Dex/Constants/Style.qml
```diff
@@ -318,6 +319,7 @@ QtObject {
                                           "IOTA": "#404040",
                                           "JSTR": "#627EEA",
                                           "DOGE": "#C3A634",
+                                          "DOI": "#C3A634",
```
smk762, line 322: Only one entry required here - you have two with different values

inspiraluna: I removed one entry and added a second commit into this PR.

PR #1588, opened 2022-02-04
Disable all no balance assets

### LINKED ISSUE #905: [FEATURE REQUEST]: Disable all zero balance coins context menu option
I have a bad habit of enabling all coins, because I'm too lazy to go thru the list and select them individually. The result of this ends up being a long list of mostly zero balance coins. The "show only coins with balance" toggle is nice, but it would be nicer to disable all the zero balance coins via a right click, similar as we have with disabling all coins of a specific type.

**Describe the solution you'd like**
Add a "disable coins with no balance" option to the right click context menu

**Describe alternatives you've considered**
Another option could be to do this via the "enable coins" form, with a checkbox that selects all zero balance coins, though this option is more clicking and less good in terms of UX.

SylEze:
Closes #905


tonymorony (approved):

smk762 (approved):
Beautiful, thanks :)

PR #1589, opened 2022-02-04
Doichain Update electrumx url

inspiraluna:
added parameter SSL to server url. added tcp protocol url.


smk762 (approved):
Thanks, loading and enabling now without issue.

tonymorony (approved):

PR #1610, opened 2022-02-09
force custom seed confirm

smk762:
closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1239

### DIFF
--- a/atomic_defi_design/Dex/Components/PopupManager.qml
+++ b/atomic_defi_design/Dex/Components/PopupManager.qml
@@ -162,9 +162,19 @@ Popup
                         field.leftPadding: dialog.isPassword ? 70 : 20
                         field.echoMode: dialog.isPassword ? TextField.Password : TextField.Normal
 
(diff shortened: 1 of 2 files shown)


tonymorony:
right now it's possible to input any text and bypass:

https://user-images.githubusercontent.com/24797699/153317163-637b94e0-28c3-440c-b35b-51dbacd38b01.mov

tonymorony (changes requested):
^^

tonymorony:
thanks @smk762 , it doesn't pass with random input now, but right now there is not good UX:

1) button OK should not be available for click if text is not match, now user have to press ok and then next to find out that his input on modal was not correct, moreover error message doesn't say it straightforward

2) input should not be case sensitive (e.g. i understand == I understand == i UnDerStAnD)

tonymorony (changes requested):
^^

tonymorony (approved):
lgtm, thank!

PR #1633, opened 2022-02-25
use full ticker for notifications

smk762:
Closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1341

### DIFF
--- a/atomic_defi_design/Dex/Constants/General.qml
+++ b/atomic_defi_design/Dex/Constants/General.qml
@@ -356,10 +356,10 @@ QtObject {
         return trail_zeros ? full_double : full_double.replace(/\.?0+$/,"")
     }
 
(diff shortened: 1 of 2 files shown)


atomic_defi_design/Dex/Constants/General.qml
```diff
@@ -356,10 +356,10 @@ QtObject {
         return trail_zeros ? full_double : full_double.replace(/\.?0+$/,"")
     }
 
-    function formatCrypto(received, amount, ticker, fiat_amount, fiat) {
```
SylEze, line 359: I think for the moment we should keep the old function for backwards compatibility and use a new one for notifications because `formatCrypto` function is used in many places inside the fontend and we need to check it didn't break anything

SylEze (changes requested):

SylEze (approved):

PR #1635, opened 2022-02-25
update NMC electrums

cipig:


smk762 (approved):
connectivity confirmed

PR #1637, opened 2022-02-25
longer, searchable wallets list

smk762:
partially solves https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1475

before:
image

after:
image

### DIFF
--- a/atomic_defi_design/Dex/Components/LinkIcon.qml
+++ b/atomic_defi_design/Dex/Components/LinkIcon.qml
@@ -9,7 +9,7 @@ Circle
     property alias source: icon.source
     property alias text: tooltip_text.text
 
-    Layout.preferredWidth: 60
+    Layout.preferredWidth: 50
     Layout.preferredHeight: Layout.preferredWidth
 
     color: Style.colorOnlyIf(mouse_area.containsMouse, Style.colorTheme4)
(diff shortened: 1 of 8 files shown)


smk762:
> can we have a searchbar please :D

https://user-images.githubusercontent.com/35845239/155859440-cd0ed3ee-1a53-4304-bd1b-cabce2957a18.mp4

howzat?

SylEze (approved):

tonymorony:
There is an excessive scroll even if wallet list fits screen

Could you please improve that? It also got sense imo show search only if list longer that screen (when scroll enabling in case of such improvement) - there is not much sense to have a search trough 2-3 elements

https://user-images.githubusercontent.com/24797699/156220331-8b54df3b-0947-4174-9662-7e5cc3dfffa4.mov

I've also discovered that at the moment we do not limit wallet name length. What can corrupt some elements of the app (wallets list included). 
Created separate issue for that: https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1640

tonymorony (changes requested):
lgtm in general, thanks! but would be great to polish a bit ^^

smk762:
> lgtm in general, thanks! but would be great to polish a bit ^^

- Search bar now only appears if wallet count > 5 
- Wallet names must be 25 char or less for new & imported wallets (closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1640)
- Excessive scroll issue should be resolved, but might need confirmation on other resolutions.

tonymorony:
the fixes looks great, thanks!

just single improvement suggestion, right now interface allows to input wallet name more than 25 chars and showing an error only when you press next button. I do propose to save user some time and just not allow to input more than 25 symbols into input field (optionally we might show the same error if she tries to input more).

image

smk762:
> the fixes looks great, thanks!
> 
> just single improvement suggestion, right now interface allows to input wallet name more than 25 chars and showing an error only when you press next button. I do propose to save user some time and just not allow to input more than 25 symbols into input field (optionally we might show the same error if she tries to input more).
> 
> image

Thanks for feedback. I've updated the wallet name validation to display error in realtime, both for when wallet name already exists and when > 25 char, and set next button to enable/disable accordingly.

SylEze (approved):

PR #1645, opened 2022-03-02
make smtf-v2 active by default on smartdex

tonymorony:
We should be extra careful on next sync to not revert such things (default coins, tos, etc)


smk762:
as the coins version has been updated, I'll await the next merge from dev, but keep this open as a reminder

smk762:
closing, moved to https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/1680

PR #1646, opened 2022-03-02
feat(protocols): add new protocols

Milerius:


smk762 (commented):
GLMR enables without issue, and price is visible.
Will need some small funds to test if transaction / swap test is needed.

smk762 (approved):
swap completed successfully

tonymorony (approved):

PR #1672, opened 2022-03-08
Addressbook: Ask user to enable disabled coin before adding an address

### LINKED ISSUE #1671: [BUG] Validation for address fails if coin not enabled
if coin not enabled, app wont let you add addressbook entry.
we can either auto-enable, or maybe like on simple view give user pop up to say  "not enabled, do you want to?"

e.g.
image

_Originally posted by @smk762 in https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1616#issuecomment-1060644493_

SylEze:
Closes #1671


smk762 (approved):
Works as intended, thanks.

PR #1697, opened 2022-03-28
Send Modal Fix: Max button does not set correct amount when unconfirmed txes

### LINKED ISSUE #1687: [BUG]: "max" button in send modal does not work if there are unconfirmed txes
image
error is because i already sent 1 LTC (out of 1.38) to another address, but that tx is not confirmed yet
i want to send the rest (0.38xx) to another address using the max button
funny is that GUI knows that i actually have only 0.38xx LTC, but the max button still tries to use 1.3x

SylEze:
Closes #1687

### DIFF
--- a/atomic_defi_design/Dex/Wallet/SendModal.qml
+++ b/atomic_defi_design/Dex/Wallet/SendModal.qml
@@ -272,7 +272,7 @@ MultipageModal
 
         property bool cryptoSendMode: true
 
-        DexRectangle
+        DefaultRectangle
         {
 
             enabled: !root.segwit && !root.is_send_busy
@@ -284,7 +284,7 @@ MultipageModal
             color: input_address.background.color
             radius: input_address.background.radius
 
-            DexTextField
+            DefaultTextField
             {
                 id: input_address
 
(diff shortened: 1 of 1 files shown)


smk762:
Looks like fee of second tx not taken into account.
image

cipig:
i get the same error when i send 1.5 LTC out of 1.99 and then try to send the rest by using max button, because it tries to send 1.99 instead of 1.99 - 1.5 = 0.49
image

smk762:
@cipig Is this replicated in CLI on your end? I've tried again recently with RICK / MORTY / LTC but had no issues.

When clicking on "max" the value input was not including the recent tx but on clicking send it reduced down to the difference.

I  wonder if it is related to https://github.com/KomodoPlatform/atomicDEX-API/issues/1254 ?

cipig:
no, only tested on GUI... if i have 5 LTC, send 2 LTC, then i need to wait for the tx with the 2 LTC to be confirmed before trying to send the rest (3 LTC)... if i don't wait, GUI will try to send 5 LTC if i click on "max", instead of 3
does GUI look at "confirmed"/"unconfirmed" answer from electrumx in result from balance? if it just takes "confirmed", it would explain this... needs to take "confirmed" + "unconfirmed" instead
but it's a small issue anyway

cipig (approved):

smk762 (approved):

PR #1734, opened 2022-04-07
Improve QRC compilation

SylEze:
How to test: just run the application and see if there aren't new QML errors and all assets are correctly loaded


smk762 (approved):
Launch test passed on windows and linux

PR #1740, opened 2022-04-15
add coins

cipig:
https://github.com/KomodoPlatform/coins/pull/462

image
image
image


smk762:
Needs to be merged in coins repo first.

smk762 (approved):

Canialon:
Interesting stuff here. I can't find new coins in an already existing wallet, but if I create a new or import one, everything will be fine.


image

cipig:
> Interesting stuff here. I can't find new coins in an already existing wallet, but everything will be fine if I create a new or import one.
> 
> coins are not added to existing wallets

That's because the settings are only imported when version changes (0.5.5 to 0.5.6) or if you manually "reset wallet configuration" from "settings"
would actually be nice, if this behaviour would be changed, so resetting the entire config is not necessary

Canialon (approved):

PR #1744, opened 2022-04-21
Decimal alignment

smk762:
closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1665
implemented in:
- simple/pro bestorders lists
- simple "from" coin selection list
- portfolio list
- transactions list.
- orderbook lists

Applied to percentages, fiat amounts and crypto amounts.

### DIFF
--- a/atomic_defi_design/Dex/Components/ColumnHeader.qml
+++ b/atomic_defi_design/Dex/Components/ColumnHeader.qml
@@ -5,7 +5,9 @@ import App 1.0
 
 Item
 {
-    property int sort_type
+    property int current_sort
+    property int sort_type: -99
+    property alias header_font: title.font
     property alias text: title.text_value
     property alias h_align: title.horizontalAlignment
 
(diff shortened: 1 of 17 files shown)


Canialon:
@cipig @smk762   I'm a little unsure about these changes. I think it looks pretty crooked. And my first impression was pretty bad.
Screenshot 2022-04-21 at 15 10 09

cipig:
i like how "Portfolio" looks now:
image

but "Best Orders" tab does not expand any more, so i can't make it to use only one line per entry when coin names are long (like NVC-QRC20):
image

can you place "CEX rate" and "Fiat price" to the right and make "You get" column expand his width in line?

cipig:
found another problem: the sorting in "Portfolio" does not work any more, i can't sort by "Change 24h", this worked before, now it looks like this:
image
same with the other columns, only sorting by "Fiat Balance" works fine

smk762:
should be solved now @cipig

cipig:
thanks, sorting is fixed in "Portfolio"

but:
image
- if width is not big enough, text can overlap in "Best Orders"
- "Price" in "Order Book" is now left-aligned, it was right-aligned before the change

smk762:
image
howzat?

cipig (approved):

Canialon (approved):

PR #1751, opened 2022-04-29
change support to modal

smk762:
closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1719


smk762:
@Canialon this should also resolve dropdown issue from https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/1749#issuecomment-1113317678

Canialon:
@smk762  Wow, making the support page look like a window, great idea, I really like it! This solution makes reading more comfortable. I thought about it for a while

****
**About max number of coins** 
I think we should create an issue about it

The app still stops responding when I select 50 and open this list again .

And another problem around 75 number. Scrolling don’t  work correctly if I choose 75:

- there are only 75, 100, 150, 200 options in the list
- if I try scroll to the top => just lagging and show other options
- if I try scroll to the bottom  => app crashing and closed

Recording:

https://user-images.githubusercontent.com/77272165/166083457-f41c9dff-f4df-4a1a-bb76-26e07e5769cd.mov

cipig:
the max coin setting problem is this: https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/1746#issuecomment-1113860463

Canialon:
Some feedback about support window:

Here the popup goes beyond the window
image
*****************

Would be great to add here some space for all of this buttons
Screenshot 2022-04-30 at 02 49 29
*****************

Should remove hover effect for this text. Because it starts to look like a button, and in a light theme it has the same color as the background.
simple (it's better to change the color to make it easier to read):
image
hover effect (omg, where is the text, haha)
image
****************

Also what do you think about  move elements a bit? :D
Something like this:
Screenshot 2022-04-30 at 03 04 19

cipig (dismissed):

Canialon:
@smk762 can you fix alignments a little bit, please?

In this screenshot, I have circled each element and center-aligned it. As you can see, they just need to be moved to the left :D
image
****************************

What do you think about making this window a little bigger in height?
************************

Also, I believe that we can remove these elements (or not?)
image

Canialon (approved):
it is perfect 🎉

Issue #1188, opened 2021-08-12
ARRR integration


artemii235:
It works in native mode only now.

1. Coins config: https://github.com/KomodoPlatform/coins/pull/303/files
2. MM2 build: latest from http://195.201.0.6/mm2.1-slp-swap/
3. It also requires http://195.201.0.6/ARRR_CACHE.db to be placed in the MM2 working directory.
4. Enable command: `curl --url "http://127.0.0.1:7783" --data "{\"userpass\":\"$userpass\",\"method\":\"enable\",\"coin\":\"ARRR\"}"`
5. No tx history is available now.
6. Block explorer link <https://pirate.explorer.dexstats.info/>
7. PIRATE daemon is required to run and be fully synced before coin activation in the GUI.

Please consider this as very early alpha/PoC. It requires the checklist from https://github.com/KomodoPlatform/atomicDEX-API/issues/927 to be implemented to avoid linking different swaps between each other, etc.

Activation time is about 20 seconds. Withdraw can take minutes depending on the transaction size.

artemii235:
Updates on the integration status:
0. Activation and withdraw are implemented using 2.0 RPC protocol https://developers.atomicdex.io/basic-docs/atomicdex-api-20-dev/#atomicdex-api-rpc-protocol-v2-0-dev. It might be required to integrate it's generic structure first.
1. Branch https://github.com/KomodoPlatform/atomicDEX-API/tree/arrr-light-mode-experiments will be usable to activate ZOMBIE and PIRATE in light mode from tomorrow.
2. Please note that even light mode requires Sapling params files have to be downloaded even for light mode (`./zcutil/fetch_params.sh/bat` from either Zcash or Komodo).
3. ZOMBIE conf:
```json
{
  "coin":"ZOMBIE",
  "asset":"ZOMBIE",
  "txversion":4,
  "overwintered":1,
  "mm2":1,
  "protocol":{
    "type":"ZHTLC",
    "protocol_data":{
      "consensus_params":{
        "overwinter_activation_height":0,
        "sapling_activation_height":1,
        "blossom_activation_height":null,
        "heartwood_activation_height":null,
        "canopy_activation_height":null,
        "coin_type":133,
        "hrp_sapling_extended_spending_key":"secret-extended-key-main",
        "hrp_sapling_extended_full_viewing_key":"zxviews",
        "hrp_sapling_payment_address":"zs",
        "b58_pubkey_address_prefix":[
          28,
[...]

PR #1775, opened 2022-05-17
Zombies

smk762:


smk762:
ref: https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1188

To enable ZHTLC coins, we first need to do `init_z_coin` which returns a `task_id`.
This is used as an input to run `init_z_coin_status` until it returns `"status": "ready"`, which will also return address / balance etc.

More info here re: coins file and mm2 version - https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1188#issuecomment-1122387546

I've been copying these into place manually after a build into /build/bin/AntaraAtomicDexAppDir/usr/share/assets/tools/mm2

To use, device also needs Sapling params files from Zcash or Komodo downloaded. Not sure if this should be done as part of app install, or if we show a pop up to direct users to do the download. 

@SylEze This is halfway there - I've setup the first step for `init_z_coin`, and current state returns below on attempt to enable

image

Now we just need to take that and complete the enable with `init_z_coin_status`.

smk762:
ZOMBIE and ARRR are marked as "wallet only" but you can remove that manually in local coins.json file to allow for swapping.
ARRR electrums are down at the moment, but I can supply some ZOMBIE (or put up a trade for you) on request for testing.

smk762:
PoC steps for delayed actions:

- when action initiates, push notification `{method} initiated. task id: {task_id}`
- when action status changes to "Ready" or "failed", push notification with result
- clicking on notification before "Ready" shows modal with current status message (e.g. sync pct enabling)

Will work towards applying the above to `init_z_coin / inti_zcoin_status` and `init_withdraw / withdraw_status` methods - @SylEze @tonymorony if you have other ideas for how to better communicate this in GUI, please let me know.

tonymorony:
@smk762 could you please sync this branch with dev?

Right now it's not building since mm2 artefact linked in it not available anymore

tonymorony:
for me nothing works atm @smk762 so even not sure from where to start.
could you please check on your side?

image

smk762:
For activation while waiting: 
- [ ] Remove from enable coins menu
- [ ] Display in portfolio and wallet page (along with enable progress)
- [ ] Notification?

smk762:
> > for me nothing works atm @smk762 so even not sure from where to start. could you please check on your side?
> 
> There appears to be an issue with the "get_balance" response parsing for ZHTLC coins so address/balance is not displaying correctly. Debugging now.

Issue is related to a misconfiguration, not sure if coins file or lightwallet_d server yet. Awaiting feedback from API team.

ZOMBIE enabling takes around 3 minutes (varies based on history tho), see: https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1188#issuecomment-1144422272

Most of this time is the `BuildingWalletDb` state, which offers no progress feedback, so we need a "please wait" message for gui users. Something like the swap progress modal might be one way to track enabling.

Proposed flow:
- user selects to enable HTLC coin.
- coin enters enabling state, no longer visible on coin activation menu.
- coin entry appears in portfolio (greyscale?) with "enabling" tag. Clicking on this coin's row opens the "Enabling progress" modal
- On wallet page, coin is visible in sidebar menu, along with "hourglass" icon. Clicking on it in the sidebar opens the "Enabling progress" modal

Currently ZHTLC enabling progress is wedged in among standard enabling, with a loop to request `task_id` status 500 times (50 was not enough!). There is potential case where this loop exits before enabling is complete (the coin will enable anyway later if no error). This loop might be better as a service, which will need to keep track of incomplete task_ids and the operation they represent (which coin, which method).

artemii235:
> Issue is related to a misconfiguration, not sure if coins file or lightwallet_d server yet. Awaiting feedback from API team.

> ZOMBIE enabling takes around 3 minutes (varies based on history tho), see: https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1188#issuecomment-1144422272

I've found a problem: Sapling activation height is ignored when blocks scanning starts, so MM2 attempts to build it from 0 block, which is not actually supported. Fixing it now.

Also, as PIRATE is quite a big chain already, it's activation time takes even more than one hour, also consuming around 2Gb of hard drive space. Since all ARRR addresses in AtomicDEX will be actually newly generated, we can choose a checkpoint block and start scanning only from it (e.g., 2000000). But if we plan to support import of already existing and long-time used addresses, users will have to wait for this scanning process to complete anyway (and it will take even longer and longer as the chain advances).

So requesting operation status only 500 times is surely not enough. This process should run until `Ready` status is received. I've also discovered that I can provide more info on `BuildingWalletDb` status about the progress. Will do it this week.

@smk762 @tonymorony

smk762:
@artemii235 could "scan_from_height" be a user configurable param? This way we could use a default "scan from" block, with option for user to override this in their coins file if needed for more mature addresses.

artemii235:
> could "scan_from_height" be a user configurable param? This way we could use a default "scan from" block, with option for user to override this in their coins file if needed for more mature addresses.

@smk762 Yes, but we will need this to be implemented from komodod side: https://github.com/KomodoPlatform/komodo/issues/558.

smk762:
500 loop was with 1 second sleep to allow closer log monitoring during testing, but I can increase the sleep to "step off" during the `BuildingWalletDb` stage so it is still monitored enough for feedback but without moving the loop counter up to quickly. Would 3 hours be a good upper limit on how long at max enabling would be expected to take?

tonymorony:
> > Issue is related to a misconfiguration, not sure if coins file or lightwallet_d server yet. Awaiting feedback from API team.
> 
> > ZOMBIE enabling takes around 3 minutes (varies based on history tho), see: [#1188 (comment)](https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1188#issuecomment-1144422272)
> 
> I've found a problem: Sapling activation height is ignored when blocks scanning starts, so MM2 attempts to build the cache from 0 block, which is not actually supported. Fixing it now.
> 
> Also, as PIRATE is quite a big chain already, its activation time takes even more than one hour, also consuming around 2Gb of hard drive space. Since all ARRR addresses in AtomicDEX will be actually newly generated, we can choose a checkpoint block and start scanning only from it (e.g., 2000000). But if we plan to support import of already existing and long-time used addresses, users will have to wait for this scanning process to complete anyway (and it will take even longer and longer as the chain advances).
> 
> So requesting operation status only 500 times is surely not enough. This process should run until `Ready` status is received. I've also discovered that I can provide more info on `BuildingWalletDb` status about the progress. Will do it this week.
> 
> @smk762 @tonymorony

Thanks for the info! 

Thats good idea with checkpoint. 
Since chain is moving I do propose not fixed height so but delta from current height e.g. 777777 blocks to the past (depends on how fast this process is to reach some satisfying UX).

And to cover other cases add configurable scan delta param to change. 
Also, maybe on GUI side we'll also need to detect if user created new seed and activating ARRR or he imported the seed and activating ARRR and treat these cases differently

tonymorony:
unattached reviewers for now
@smk762 please re-request review when it'll be possible to send/receive/trade ARRR and see tx history (even if it'll look not fancy / POC)

src/core/atomicdex/services/mm2/mm2.service.cpp
```diff
@@ -702,11 +740,134 @@ namespace atomic_dex
...
+
+                                                    SPDLOG_DEBUG("{} init_z_coin Task ID: {}", tickers[idx], task_id);
+
+                                                    nlohmann::json j = ::mm2::api::template_request("init_z_coin_status", true);
+                                                    ::mm2::api::to_json(j, z_request);
+                                                    z_batch_array.push_back(j);
+                                                    std::string last_event = "none";
+                                                    std::string event = "none";
+
+                                                    // set to enabled "early" so it shows up in models
```
smk762, line 766: Uncommenting this will show the coin in wallet pages, but blocks the login or enable screen which waiting.
Commenting it out, coin will begin to enable but wont be removed from enable menu or added to wallet pages.

Because `currently_enabled is boolean, we'll need an extra state here like `enabling_in_progress`.

artemii235:
> Since chain is moving I do propose not fixed height so but delta from current height e.g. 777777 blocks to the past (depends on how fast this process is to reach some satisfying UX).

@tonymorony 
If we release at e.g. 2000000 block, at some point `current - delta` will be ahead of 2000000 and some users might stop seeing correct balance if they HODL for all this time. Also, 777777 is too large, wallet DB generation takes like 8-9 seconds per 1000 blocks. I have to wait for over 15 minutes even with 1900000 block as checkpoint (and I have no transactions on the address). 

@smk762 
I have added check point feature and more info to BuildingWalletDb status, please try adding the following to the PIRATE conf:
https://github.com/KomodoPlatform/atomicDEX-API/blob/626624c96c5469fabc7cd6edf49b131aa1cd9148/mm2src/mm2_test_helpers/src/for_tests.rs#L216

`BuildingWalletDbStatus` status is changed to:
```json
{
  "mmrpc":"2.0",
  "result":{
    "status":"InProgress",
    "details":{
      "BuildingWalletDb":{
        "current_scanned_block":1000,
        "latest_block":215326
      }
    }
  },
  "id":null
}
```

artemii235:
>  I have to wait for over 15 minutes even with 1900000 block as checkpoint (and I have no transactions on the address).

I will also check if this can be optimized.

artemii235:
The code is in the https://github.com/KomodoPlatform/atomicDEX-API/tree/arrr-activation-hotfix branch. CI will publish binaries to http://195.201.0.6/ soon.

smk762:
Thanks @artemii235 I can enable ARRR now. Will test it further this weekend.

@SylEze here is a visual indication of progress (and what's missing)

https://user-images.githubusercontent.com/35845239/183118936-21104473-ad8c-4eb0-a446-9b6074243a66.mp4

- ZHTLC Coins can be enabled, but it takes a while.
- Activation progress is tracked in logs, but we need to do it for front end too (e.g. hourglass next to coin in portfolio, click on to get progressbar / event). 
- Currently while activating, coin stays on enable menu and doesn't appear in portfolio. This is governed by the boolean `m_coins_informations[tickers[idx]].currently_enabled`. We need to add a third state which covers "enabling in progress" so enable menu and portfolio display is sane.
- Transaction history display is implemented, though I have seen some errors on one of my testing wallets with a bit more history `error: Response(zombie.sirseven.me:10033, Object({\"code\": Number(-32600), \"message\": String(\"response too large (over 1,000,000 bytes\")}))` so we need to handle this.

https://user-images.githubusercontent.com/35845239/183121329-3f332af0-2786-4234-ae55-39be28079cfd.mp4

image

image

- Sending funds is implemented.
- Swaps should be working, but I have not re-confirmed since re-aligning code with dev. I need to source some ARRR, but I can send you some ZOMBIE for testing if needed, just DM me an address.

@tonymorony The current coin icon image for zombie is just a random placeholder I found online somewhere. Can our artists create something? Or could we have a community competition / bounty for creating / selecting the image?

smk762:
I've imported and merged in @Milerius 's SLP code and enabling is now functional - but is failing to enable tokens if BCH/tBCH is not already enabled. So we need to handle the same way as other platform parent / token enabling logic.

For testing, here is a working faucet: https://tbch.googol.cash/ or from a telegram bot at https://t.me/tbch_bot
SLP token explorer: https://simpleledger.info/

artemii235:
> Transaction history display is implemented, though I have seen some errors on one of my testing wallets with a bit more history error: Response(zombie.sirseven.me:10033, Object({\"code\": Number(-32600), \"message\": String(\"response too large (over 1,000,000 bytes\")})) so we need to handle this.

I think the easiest workaround is to increase it in Electrum's config @tonymorony @SirSevenG 

But since `z_coin_tx_history` requests transaction data from electrums in batch, certain response size will be surely reached at the certain limit. However, I wouldn't recommend to use big `limit` for any of calls in general. It can allocate a lot of RAM, which might not be freed for entire MM2 execution.

smk762:
fwiw, this only happened on my test wallet with lots of tx - reduced limit to 50 in gui code and it is ok now (though other coins seem to handle more). Max I got returned without error was around 90.

artemii235:
ARRR activation fix, `TemporaryError` RPC status and check point block feature is released https://github.com/KomodoPlatform/atomicDEX-API/releases/tag/beta-2.1.7404 @smk762 @tonymorony

tonymorony:
Wow, great progress @smk762 !

Noticed minor issue:

Screenshot 2022-08-24 at 12 12 20

Opening such link: 

https://pirate.explorer.dexstats.info/address/Invalid%20Ticker

It would be great to not show this link for pirate

SirSevenG:
There seems to be no option to generate/get zcash params in app.
So trying to activate coins leads to error:
```

{
    "id": null,
    "mmrpc": "2.0",
    "result": {
        "details": {
            "error": "Error on platform coin ZOMBIE creation: ZCashParamsNotFound",
            "error_data": {
                "error": "ZCashParamsNotFound",
                "ticker": "ZOMBIE"
            },
            "error_path": "lib.z_coin_activation.z_coin",
            "error_trace": "lib:93] z_coin_activation:191] z_coin:781]",
            "error_type": "CoinCreationError"
        },
        "status": "Ready"
    }
}
```
I don't see most users using Adex-Desktop on the same device running full node, so that will be a big issue.

tonymorony:
Nice catch @SirSevenG :)
Most of the users never setup a full node so it's a blocker and we should detect this error and handle for the user params downloading
(prob we can just execute https://github.com/KomodoPlatform/komodo/blob/master/zcutil/fetch-params-alt.bat and https://github.com/KomodoPlatform/komodo/blob/master/zcutil/fetch-params-alt.sh)

tonymorony (changes requested):
^^

tonymorony (dismissed):
I was able to send/receive/swap ARRR and zcash params downloaded fine. Awesome work @smk762 🙏 !
I've tested only on OSX so lets wait for results of QA multios test @Canialon @SirSevenG and @SylEze 's code review b4 merging

Canialon (changes requested):
Zombie page has a faucet button 🤔 
image

____________
Can you please add auto-enable assets that which user selected after downloading Zcash ?)
Because now it looks like:
user selected ARRR/Zombie -> opened window about Zcash downloading -> user downloaded it -> no coins activated -> user need to select them again

smk762:
> Zombie page has a faucet button thinking image

I've setup a couple of Zombie nodes to send faucet transactions from, and modified the faucet code to handle ZOMBIE with the alternative `z_` methods for sending and address validation. 

Tested myself and should be working. I can't check the balances of requesting addresses tho, so individual rate limiting is more difficult. Fortunately I can mine 100k zombie overnight :laughing:

tonymorony:
`There are 2 issues with zparams download i think should be addressed:`

Partial downloading support is not trivial thing and overkill for this task imo

smk762:
I've since confirmed with Artem also that the sprout params are not needed, which significantly reduces download time.

smk762:
> There are 2 issues with zparams download i think should be addressed:
> 
> 1. Continue download if interrupted

This was not implemented (ref Tony's comment), but I've removed the sprout params from download so it's really quick.

> * try to activate any of Z-coins
> * See zparams not found error
> * Nothing happens
> * try to activate any of Z-coins again
> * Popup for downloading zparams opens
> * Start download
> 
> It's quite confusing for the user. )

I think what happened here might be a click outside the modal as it was loading, so it auto-closed. I've added `NoAutoClose` policy to the modal, so now it will only close if use clicks the close button, or once the download completes.

On completion of downloads, modal automatically closes, and previously selected ZHTLC coins start the activation process again.

Canialon:
Quick new moment:
ARRR and Zombie received transaction details don’t display "from" address
Screenshot 2022-09-03 at 05 34 20

smk762:
> Quick new moment: ARRR and Zombie received transaction details don’t display "from" address 
I see both addresses in all my transactions in 3 different wallets, except the very first tx to one of the address.
Can you pls confirm how many tx you see this with? Will try and track it down, seems an edge case and maybe a "feature" of the coin.

SirSevenG:
> > Quick new moment: ARRR and Zombie received transaction details don’t display "from" address
> 
> I see both addresses in all my transactions in 3 different wallets, except the very first tx to one of the address. Can you pls confirm how many tx you see this with? Will try and track it down, seems an edge case and maybe a "feature" of the coin.

I think it's chain quirk with z-transactions. AE: i can see From on my own transaction, but can't see one from faucet transaction.
Also explorer doesn't help much :D https://zombie.explorer.lordofthechains.com/tx/a169bd5133209921d332ad4243d6fd3170038deafbe7d92c182a05fbc933ac24

tonymorony:
thats actually the main PIRATE feature :)

SirSevenG (dismissed):
Yep, own transaction from full node:
image

Unless we want to hide "From" field on such transactions or add smth like "Shielded adress" to the field, i see no issues with this PR now.
Was able to enable, swap and transact with Zombie and Pirate, faucet works fine for me. There's a small issue with ZOMBIE mining speed, but that's not related. :D

tonymorony (dismissed):

Canialon:
In general, this pr looks good to me^^ Thanks!
Except for this faucet window, the text doesn’t fit and overflows. I'm unsure if we need to fix it in this pr or create a separate one.

image

smk762:
> In general, this pr looks good to me^^ Thanks! Except for this faucet window, the text doesn’t fit and overflows. I'm unsure if we need to fix it in this pr or create a separate one.
> 
> image

Thanks for finding - I've updated the faucet to make the message shorter and it should fit now.

assets/config/0.5.6-coins.json
```diff
@@ -10698,5 +10800,45 @@
...
+    "type": "ZHTLC",
+    "active": false,
+    "currently_enabled": false,
+    "wallet_only": false
+  },
+  "ARRR": {
+    "coin": "ARRR",
+    "name": "Pirate",
+    "nomics_id": "ARRR",
+    "coinpaprika_id": "arrr-coin",
```
cipig, line 10828: id is `arrr-pirate`, like in https://coinpaprika.com/coin/arrr-pirate/

cipig:
image
orderbook shows wrong info... in this case it shows an ask of 210 PND, but the order is with 270559 PND... when i click on that ask the volume in order placing gets prefilled with 270559, so the correct value is present... idk where 210 comes from

cipig:
ask is now fixed, but bid is wrong
image
it should be 2 KMD for 11554 PND, like in the output from `curl -s --url "http://127.0.0.1:7783" --data "{\"mmrpc\":\"2.0\",\"userpass\":\"$userpass\",\"method\":\"orderbook\",\"params\": {\"base\":\"PND\",\"rel\":\"KMD\"}}"`:
```
    "base": "PND",
    "bids": [
      {
        "coin": "KMD",
        "address": {
          "address_type": "Transparent",
          "address_data": "RT9MpMyucqXiX8bZLimXBnrrn2ofmdGNKd"
        },
        "price": {
          "decimal": "0.0001731002250302925393803011943915527090185217240782413017136922277998961398649818244763718192833650684"
        },
        "pubkey": "031bb83b58ec130e28e0a6d5d2acf2eb01b0d3f1670e021d47d31db8a858219da8",
        "uuid": "bee33e59-eb29-4f0f-8286-dcf04cc1b38d",
        "is_mine": false,
        "base_max_volume": {
          "decimal": "11554"
        },
        "base_min_volume": {
          "decimal": "10"
        },
        "rel_max_volume": {
          "decimal": "2"
        },
        "rel_min_volume": {
          "decimal": "0.001731002250302925393803011943915527090185217240782413017136922277998961398649818244763718192833650684"
        },
        "conf_settings": {
          "base_confs": 1,
          "base_nota": false,
          "rel_confs": 4,
          "rel_nota": false
        },
        "base_max_volume_aggr": {
          "decimal": "11554"
        },
        "rel_max_volume_aggr": {
          "decimal": "2"
        }
      }
    ],
```

cipig:
orderbook is fixed, thanks
but best_orders are not showing up, instead there is an error in mm2.log related to it (shows up every time i click on an order in orderbook)
```
07 00:55:53, rpc:173] ERROR RPC dispatcher error: dispatcher:94] Error parsing request: missing field `request_by`
```

cipig (dismissed):

Canialon (approved):

Canialon:
Noticed just now, zombie colourful name is not visible very good on a dark theme. Can you change the color of it somehow?
image

tonymorony (approved):

PR #1777, opened 2022-05-23
Remake app update checker

### LINKED ISSUE #1228: [FEATURE REQUEST]: Application updater polishing
Let's combine the previous version checker we had (komodo.live/adexproversion) with auto-updater, with the actual version number/link specifying on the endpoint, so:

a) if something will go wrong as for win 0.5.0 we'll be able to change the version number on the endpoint to old and users will not see notifications
b) if GH will change API structure we'll able just update the link on endpoint
c) we'll be able to block app login if a hard forking update needs to be installed (as it was before with 3 states: availiable/required/mandatory)

So same view/notifications as we had before with komodo.live endpoint just with button to auto-update and info from endpoint

SylEze:
Closes #1228


SylEze:
(cc. @Canialon @smk762)

smk762 (approved):
Tested and sighted on all OS. A couple of times with installers, did not see on first launch, but was visible on next launch.
Current update link points to older version 0.4.2. (assume will be resolved prior to release)

PR #1785, opened 2022-06-06
Change update modal behaviour if the update is recommended or mandatory

SylEze:
Auto open update modal if update is recommended or mandatory
Disallow using app if update is mandatory

How to test:
- Change values sent by the following POST API: https://komodo.live/adexproversion?currentVersion=0.5.5
- If "status" has been changed to "mandatory" or "recommended", Dex should opens the new update modal itself
- Also if "status" is "mandatory", you cannot do anything other than opening the download link or closing the application

### DIFF
--- a/atomic_defi_design/Dex/NewUpdateModal.qml
+++ b/atomic_defi_design/Dex/NewUpdateModal.qml
@@ -1,5 +1,7 @@
 import QtQuick 2.12
 import QtQuick.Layouts 1.12
+import QtQml 2.15               //> Qt.exit
+import QtQuick.Controls 2.15    //> Popup.NoAutoClose
 
 import "Components" as Dex
 import "Constants" as Dex
@@ -18,6 +20,17 @@ Dex.MultipageModal
             return 2
         return 3
     }
+
+    Component.onCompleted:
+    {
+        if (Dex.API.app.updateCheckerService.updateInfo.status === "recommended" ||
+            Dex.API.app.updateCheckerService.updateInfo.status === "required" )
+        {
+            root.open()
+        }
+    }
+
+    closePolicy: Popup.NoAutoClose
 
     Dex.MultipageModalContent
     {
@@ -72,7 +85,7 @@ Dex.MultipageModal
 
     Dex.MultipageModalContent
     {
-        titleText: qsTr("New version found")
+        titleText: Dex.API.app.updateCheckerService.updateInfo.status === "required" ? qsTr("Mandatory version found") : qsTr("New version found")
         titleAlignment: Qt.AlignHCenter
         spacing: 16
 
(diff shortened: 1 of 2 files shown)


smk762 (approved):

SylEze:
> After reviewing this I noticed that when update is mandatory, you can "bypass". When clicking on download, modal goes away (and webpage loads) and I can log into old version
> bypass.mp4
> 
> .

Oopsie 😅

smk762:
endpoint currently returns status `available`:
```[20:22:28] [debug] [update.checker.service.cpp:71] [2434259]: App update service status is: available
[20:22:28] [debug] [update.checker.service.cpp:72] [2434259]: App update service currentVersion is: 0.5.5
[*** LOG ERROR #0001 ***] [2022-06-19 20:22:28] [log_mt] {[json.exception.type_error.302] type must be string, but is boolean}
[20:22:28] [debug] [update.checker.service.cpp:74] [2434259]: App update service newVersion is: 0.5.6
[20:22:28] [debug] [update.checker.service.cpp:75] [2434259]: App update service downloadUrl is: https://github.com/KomodoPlatform/atomicDEX-Desktop/releases/tag/0.4.2-beta
```
With this status, I get no update pop up on loading. If clicking the "search update" button in settings, I get
image

When hardcoding status as "recommended" I get the same screen as in image above. The "Close Dex" button does not close the app.

When hardcoding status as  "required"
image
App blocks as expected, can't bypass, and "close dex" button works as advertised.

SylEze:
> When hardcoding status as "recommended" I get the same screen as in image above. The "Close Dex" button does not close the app.

Yeah recommended updates should not force to close application, the text was just wrong and I just fixed it

atomic_defi_design/Dex/NewUpdateModal.qml
```diff
@@ -19,6 +21,17 @@ Dex.MultipageModal
         return 3
     }
 
+    Component.onCompleted:
+    {
+        if (Dex.API.app.updateCheckerService.updateInfo.status === "recommended" ||
+            Dex.API.app.updateCheckerService.updateInfo.status === "required" )
+        {
+            root.open()
+        }
```
smk762, line 30: Can we add else case to add to logs if status is ok or api returns status_code != 200?
```
        else
        {
            console.log(Dex.API.app.updateCheckerService.updateInfo.status)
        }
```

SylEze: Yes ofc

smk762: I tried this locally and seems would need to add to logs on back end not front.

smk762 (approved):
Mandatory case (no bypass) :heavy_check_mark: 
Optional case :heavy_check_mark: 
API down case: :heavy_check_mark: (no error or popup shown)
No update case: :heavy_check_mark: (no error or popup shown)

PR #1873, opened 2022-07-15
Prompt for password when disabling 2FA / privacy mode

smk762:
closes: https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1862

### DIFF
--- a/atomic_defi_design/Dex/Components/PopupManager.qml
+++ b/atomic_defi_design/Dex/Components/PopupManager.qml
@@ -166,12 +166,15 @@ Popup
 
                         field.onTextChanged:
                         {
-                            if (validator(field.text))
+                            if (validator)
                             {
-                                dialog.enableAcceptButton = true
-                            }
-                            else {
-                                dialog.enableAcceptButton = false
+                                if (validator(field.text))
+                                {
+                                    dialog.enableAcceptButton = true
+                                }
+                                else {
+                                    dialog.enableAcceptButton = false
+                                }
                             }
                         }
                         field.onAccepted:
--- a/atomic_defi_design/Dex/Settings/SettingModal.qml
+++ b/atomic_defi_design/Dex/Settings/SettingModal.qml
@@ -457,6 +457,7 @@ Qaterial.Dialog
                                 width: parent.width - 30
                                 anchors.horizontalCenter: parent.horizontalCenter
                                 height: 60
+
                                 DexLabel
                                 {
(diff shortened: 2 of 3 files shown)


Canialon:
1. enable 2FA
2. try to disable 2FA -> appeared the window to enter the password 
3. enter the correct password 

Result: 
- the same "enter password" window appeared
- 2FA was not disabled

**********************

I guess it would be fine to specify which password the user needs to enter (wallet or system). Because 2FA asks for the system password, so I was confused about it
image

Canialon (changes requested):
^

SirSevenG (dismissed):

Canialon (changes requested):
1. enable 2FA
2. click to disable 2FA protection 
3. do NOT touch anything 

Actual result:
- appeared "disable 2FA" window
- 2FA already disabled 💀 🎉 (without password confirmation and it is no reason what you will press on the confirmation window)

Expected result:
 - 2FA disabled only after password confirmation

tested on MacOS and Windows

smk762:
> 1. enable 2FA
> 2. click to disable 2FA protection
> 3. do NOT touch anything
> 
> Actual result:
> 
> * appeared "disable 2FA" window
> * 2FA already disabled skull tada (without password confirmation and it is no reason what you will press on the confirmation window)
> 
> Expected result:
> 
> * 2FA disabled only after password confirmation
> 
> tested on MacOS and Windows

Visually in the background, the toggle will go to "off", but if no / bad password it will exit modal and turn on again (it stays on in the backend, this is just visual).
I did notice that clicking "cancel" instead of entering password did not turn the toggle back on though - this is now fixed.

Canialon:
> I did notice that clicking "cancel" instead of entering password did not turn the toggle back on though - this is now fixed.

omg, I forgot to write this in my report🙈 thank you for noticing! 
____________

In general, I'm ready to approve this PR coz functionality of disabling 2FA works well now. 
But it would be fine to improve one moment if we have time.
1. enable 2FA
2. click to disable 2FA protection
3. enter incorrectly password -> appeared "wrong password" window
image
This window includes two buttons, but they have the same functionality. It would be fine to remove one button or rather add red borders to the password field if the password is wrong.
image

Because now, users should:
1. enter the wrong password
2. close the "wrong password" window
3. click on the "disable 2FA" to try again

So users haven’t a chance to try to enter the password immediately.

tonymorony (approved):

PR #1883, opened 2022-07-21
avoid 2 line crypto amount

smk762:
closes: https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1844

Where a ticker is too long to show without using 2 lines, only the numeric value will be shown.


Canialon (approved):
a good solution:) thanks!

tonymorony (approved):

PR #1889, opened 2022-07-27
Fix macos notarise

tonymorony:


tonymorony:
@smk762 was a deep dive lol, notarisations should be fixed now for both dmg and installer (it looks like apple added team indeficator parameter ASC_PUBLIC_ID as mandatory just recently and macos_notarize.sh not updated nor in our copypaste, nor in original tool https://github.com/drud/signing_tools/blob/master/macos_notarize.sh

smk762 (approved):
Nice one )
Working as expected on mac here with latest CI binaries.

PR #1891, opened 2022-07-28
Use Actions Environments

smk762:


smk762:
This turns env vars into secrets and the ones we're using here dont need to be hidden so it is simpler to define in the actual yaml

PR #1895, opened 2022-07-29
Beautify address validation error message

smk762:
Closes: https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1786

To test, try to enter a KMD address for a BTC entry in the addressbook.

### DIFF
--- a/atomic_defi_design/Dex/Constants/General.qml
+++ b/atomic_defi_design/Dex/Constants/General.qml
@@ -150,6 +150,14 @@ QtObject {
         }
     }
 
+    function clean_warnings(text) {
+        if (text.includes("Backend error")) {
+            pos = text.lastIndexOf("]") + 1
+            return text.substring(pos).trim()
+        }
+        return text
+    }
+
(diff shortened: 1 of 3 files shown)


atomic_defi_design/Dex/Constants/General.qml
```diff
@@ -150,6 +150,14 @@ QtObject {
         }
     }
 
+    function clean_warnings(text) {
```
SylEze, line 153: I think this one should be done via the backend:
https://github.com/KomodoPlatform/atomicDEX-Desktop/blob/dev/src/core/atomicdex/pages/qt.wallet.page.cpp#L329

SylEze (changes requested):

atomic_defi_design/Dex/Constants/General.qml
smk762: thanks for the tip - updated :+1:

SylEze (approved):

tonymorony (approved):

PR #1937, opened 2022-08-15
UTXO merging

smk762:
Related to https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1784

Currently this is only manually configurable by adding `"utxo_merge": true` to a coins config (e.g. in 0.5.6-coins-username.json)
It will only work for UTXO coins.

In future, we can allow this to be configured thru the gui settings.

@gcharang - hopefully this will make explaining how to recover an address from bulk mining utxos a little simpler to the non-technical users.

### DIFF
--- a/src/core/atomicdex/api/mm2/rpc.electrum.cpp
+++ b/src/core/atomicdex/api/mm2/rpc.electrum.cpp
@@ -35,10 +35,12 @@ namespace mm2::api
             j["swap_contract_address"] = cfg.is_testnet ? cfg.testnet_qrc_swap_contract_address : cfg.mainnet_qrc_swap_contract_address;
             j["fallback_swap_contract"] = cfg.is_testnet ? cfg.testnet_fallback_qrc_swap_contract_address : cfg.mainnet_fallback_qrc_swap_contract_address;
         }
-        if (cfg.address_format.has_value()) {
+        if (cfg.address_format.has_value())
+        {
             j["address_format"] = cfg.address_format.value();
         }
-        if (cfg.merge_params.has_value()) {
+        if (cfg.merge_params.has_value())
+        {
             j["utxo_merge_params"] = cfg.merge_params.value();
         }
     }
--- a/src/core/atomicdex/api/mm2/rpc.electrum.cpp
+++ b/src/core/atomicdex/api/mm2/rpc.electrum.cpp
@@ -41,7 +41,6 @@ namespace mm2::api
         if (cfg.merge_params.has_value()) {
             j["utxo_merge_params"] = cfg.merge_params.value();
         }
-        SPDLOG_INFO("electrum: {}", j.dump());
     }
 
     //! Deserialization
--- a/src/core/atomicdex/api/mm2/rpc.electrum.cpp
+++ b/src/core/atomicdex/api/mm2/rpc.electrum.cpp
@@ -38,7 +38,10 @@ namespace mm2::api
         if (cfg.address_format.has_value()) {
             j["address_format"] = cfg.address_format.value();
         }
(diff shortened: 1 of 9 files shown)


src/core/atomicdex/services/mm2/mm2.service.cpp
```diff
@@ -641,8 +641,17 @@ namespace atomic_dex
                     request.address_format                   = nlohmann::json::object();
                     request.address_format.value()["format"] = "segwit";
                 }
+
+                if (coin_info.utxo_merge.value_or(false))
+                {
+                    t_utxo_merge_params params{.merge_at = 300, .check_every = 60, .max_merge_at_once = 200};
```
cipig, line 647: we would have to test if it works with max_merge_at_once=200, i can imagine it would cause issues with some electrums, so maybe it's better to set max_merge_at_once to 100
check_every = 60 means it checks every minute... do we need to do this every minute? maybe better to do it every 10 minutes or so
any maybe also better to start earlier, not wait for 300 utxos
what do you guys think?

smk762: I"m happy to tweak the default params - just used those to drop around 3500 RICK utxos into 130 or so relatively quickly.
There is a user with 23000 utxos from mining tokel, which will take a while to consolidate either way.

From memory, up to 800 or so utxos can be merged in a tx.
I"m not sure at what number utxos electrums begin to complain about "payload too large"
I think every 5 min is a balance between not too obtrusive, and not too slow.

cipig: OK, sounds like it will work... there is a difference though, RICK (and all other coins run by me) electrums have `MAX_SEND = 2000000` set, the default is 1M

src/core/atomicdex/api/mm2/rpc.electrum.cpp
```diff
@@ -38,7 +38,9 @@ namespace mm2::api
         if (cfg.address_format.has_value()) {
             j["address_format"] = cfg.address_format.value();
         }
-        //SPDLOG_INFO("electrum: {}", j.dump());
+        if (cfg.merge_params.has_value()) {
```
SylEze, line 41: Allman style indentation

src/core/atomicdex/services/mm2/mm2.service.cpp
```diff
@@ -641,8 +641,17 @@ namespace atomic_dex
...
+                if (coin_info.utxo_merge.value_or(false))
+                {
+                    t_utxo_merge_params params{.merge_at = 250, .check_every = 300, .max_merge_at_once = 125};
+                    nlohmann::json             j;
+                    mm2::api::to_json(j, params);
+                    request.merge_params     = j;
+                }
                 nlohmann::json j = ::mm2::api::template_request("electrum");
                 ::mm2::api::to_json(j, request);
+                // SPDLOG_INFO("electrum request: {}", j.dump(4));
```
SylEze, line 654: remove

SylEze (changes requested):

src/core/atomicdex/services/mm2/mm2.service.cpp
smk762: updated to use these values:
merge_at = 250
check_every = 300
max_merge_at_once = 125

SirSevenG (approved):

tonymorony:
@smk762 it would be awesome do add text notification about merging event so user will see not just a notification about transaction (what might be scary/suspicious since user not initiated it) but also a message why this transaction happened

tonymorony (approved):

PR #1962, opened 2022-09-09
Sidebar tweak

smk762:
Closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1828
@pbca26 does this solution work for you?

https://user-images.githubusercontent.com/35845239/189441936-8f789d69-141c-44fe-b97c-05ebd7982385.mp4

### DIFF
--- a/assets/config/cfg.json
+++ b/assets/config/cfg.json
@@ -1,5 +1,6 @@
 {
   "notification_enabled": true,
+  "sidebar_enabled": true,
   "current_currency": "USD",
   "current_fiat": "USD",
   "current_currency_sign": "$",
--- a/atomic_defi_design/Dex/Screens/Dashboard.qml
+++ b/atomic_defi_design/Dex/Screens/Dashboard.qml
@@ -105,9 +105,9 @@ Item
     // Right side
     AnimatedRectangle
     {
-        width: parent.width - sidebar.width
+        width: parent.width - 80
         height: parent.height
-        x: sidebar.width
+        x: 80
         border.color: 'transparent'
 
         // Modals
@@ -265,7 +265,6 @@ Item
     Sidebar.Main
     {
         id: sidebar
-
         enabled: loader.status === Loader.Ready
 
         onLineSelected: currentPage = lineType;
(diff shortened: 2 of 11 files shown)


Canialon (changes requested):
@smk762 as for me, solution to add an option to the settings is very good. Sounds nice to have opportunity to manage this part. This sidebar can be quite annoying, but if user already know all tabs he/she can just turn off animation of sidebar opening :/ It is possible?
And just leave hover effect with name of hovered tab if sidebar is off

also logo is jerky in moment of switching between big and little variants, can you make it animated? Maybe growing idk

ps: I just found few issues because of sidebar haha, but this pr will leave no chance to them \*sound of crying bugs*

smk762:
> @smk762 as for me, solution to add an option to the settings is very good. Sounds nice to have opportunity to manage this part. This sidebar can be quite annoying, but if user already know all tabs he/she can just turn off animation of sidebar opening :/ It is possible? And just leave hover effect with name of hovered tab if sidebar is off
> 
> also logo is jerky in moment of switching between big and little variants, can you make it animated? Maybe growing idk
> 
> ps: I just found few issues because of sidebar haha, but this pr will leave no chance to them *sound on crying bugs*

I've added a toogle. Tooltips are a bit tricky here - we used to have them but positioning anywhere except above is a challenge, and in this case it goes over the icon above. I think users turning off the sidebar expansion will know what they are without a tooltip.
Enlarging the logo as it expands is also a bit of a challenge because it's two different images and managing size & alignment to sync is temperamental. I tweaked it a bit tho, so the two images overlay seamlessly at the end of the transition.

Canialon (changes requested):
Thanks a lot, @smk762 !
But the setting can crash D:

1. log in to any wallet
2. enable the "disable expanding sidebar" setting (to make sidebar static)
3. log in to another wallet
4. try to use the buttons in sidebar
-> the hover effect is broken, icons of the tab stay blue even if a user opens another tab
5. try to disable the "disable expanding sidebar" setting 
-> it is not working, the sidebar is still static and with blue icons :(

To fix: reset wallet.

I was able to reproduce it on windows and mac

smk762:
Thanks for feedback @Canialon :pray: 
I've put this back to draft and will approach it again in 0.5.8.

smk762:
related - https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2051

smk762:
Closing, sidebar rework needs more discussion to determine which option will be used.

PR #1964, opened 2022-09-11
add tamadoge

smk762:
coins repo ref: https://github.com/KomodoPlatform/coins/pull/488


cipig (approved):

tonymorony (commented):
shouldn't there be changes in QML file?

cipig:
> shouldn't there be changes in QML file?

you mean for the icon? There is no QML file any more.
Only a color definition in atomic_defi_design/Dex/Constants/Style.qml is missing... but if there isn't one, a default color will be used.

smk762:
> shouldn't there be changes in QML file?

It reverts to a standard color for theme if not defined. I'm not sure adding custom color def for each coin is so efficient - might be better to do it based on protocol / platform.

https://github.com/KomodoPlatform/atomicDEX-Desktop/blob/c1a5f19bde94e70f099fe2aaf64d6c086453a9a2/atomic_defi_design/Dex/Constants/Style.qml#L223-L226

tonymorony (approved):

PR #1967, opened 2022-09-13
restore default coins json if user json invalid

smk762:
Closes: https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1966

To test: 
- [ ] Edit your coins json file so it is invalid (e.g. delete some `"` or `,` or `}`)
- [ ] Launch app, should be able to log in without crash, and see default coins enabled.


smk762:
closing, combined with https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/1632

smk762:
this is probably worth having in next release even if https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/1632 is not ready yet.

SirSevenG:
Works for me.
Broke coins-config file, app started with default config for the wallet with no issues.

Not sure if we should notify user why their coins list was reset. This is an unlikely case for normal app updates.

cipig (approved):

PR #1982, opened 2022-09-26
update coins file, fix logic

smk762:
Thanks @cipig for reporting BCH no longer enabling in dev.
Once https://github.com/KomodoPlatform/coins/pull/493 is merged, I'll rerun CI to test this.

- [ ] Can enable tBCH & BCH
- [ ] Can enable SLP tokens

To test locally, you can change https://github.com/KomodoPlatform/atomicDEX-Desktop/blob/dev/CMakeLists.txt#L82 
and use `https://github.com/KomodoPlatform/coins/archive/slptokens.zip`


SylEze:
I guess the two coins files conflict ? Take the one with the "other_types" field for BCH/tBCH please

PR #1991, opened 2022-09-29
fix chart from bestorders bug

smk762:
Closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/1915
To test:
- Load app, go to pro dex view. Select order in best orders in sell mode, confirm correct chart loads
- Select order in best orders in buy mode, confirm correct chart loads
- change right combo coin selected, confirm correct chart loads
- change left combo coin selected, confirm correct chart loads
- flip left / right coins with icon between combos, confirm correct chart loads


cipig:
switching pairs to best_orders (buy and sell) fixed
but if you switch pairs by changing the left coin through dropdown, the chart is inverted, shows eg KMD/BCH instead of BCH/KMD

PR #1996, opened 2022-10-01
show activated if activated

smk762:
Fixes a bug where you can't enable a coin because when you do the API response in logs is 
```
[20:48:23] [debug] [mm2.service.cpp:1049] [922060]: error: bad answer json for enable/electrum details: {
    "error": "rpc:212] dispatcher_legacy:160] lp_commands_legacy:121] lp_coins:2270] Coin ZRX-PLG20 already initialized"
}
```
In this case, PR will make sure app adds coin to be visible in portfoilio.
To test:
- launch app, and get rpc password from `settings > security > view seed and private key`
- use rpc password to [activate coins in CLI](https://developers.komodoplatform.com/basic-docs/atomicdex-api-legacy/coin_activation.html). To cover all cases, include at least one of each: UTXO / ERC-20 / SLP / ZHTLC
- Coins will not be visible in app. Select them for enabling from coin activation menu.
- Confirm coin is now visible in portfolio, and when you go to wallet page and click "Receive" it shows correct address and not `Invalid Ticker` (you might see this if you are fast enough, but address should appear within 3-5 sec at most)

### DIFF
--- a/atomic_defi_design/Dex/Dashboard/NotificationsModal.qml
+++ b/atomic_defi_design/Dex/Dashboard/NotificationsModal.qml
@@ -242,6 +242,8 @@ DexPopup
         const change = General.formatFullCrypto("", amount, ticker, "", "", true)
         if (!app.segwit_on)
         {
+            if (amount != 0)
+            {
             newNotification("onBalanceUpdateStatus",
                 {
                     am_i_sender,
@@ -255,6 +257,7 @@ DexPopup
                 qsTr("Your wallet balance changed"),
                 human_date,
                 "open_wallet_page")
+            }
         }
         else
         {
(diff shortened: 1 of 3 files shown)


SirSevenG (changes requested):
On Ubuntu 20:
Started App
Enabled RICK in cli
Activated RICK in application
Opened RICK in dashboard, pressed faucet button
Application crashed

I don't see anything in logs tho, but it's reproducible scenario.

smk762:
> On Ubuntu 20: Started App Enabled RICK in cli Activated RICK in application Opened RICK in dashboard, pressed faucet button Application crashed
> 
> I don't see anything in logs tho, but it's reproducible scenario.

Thanks - the cause of the crash was the RICK address not yet being set correctly (showing as `invalid ticker`) and not completely enabled.

SirSevenG (approved):

smk762:
I've added some simple AtomicDEX-API bash scripts @ https://github.com/smk762/DragonhoundTools/tree/master/atomicdex/bash to assist with testing in CLI for PRs like this.

Canialon (approved):

PR #2028, opened 2022-10-27
fix custom gas

smk762:
closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2019

### DIFF
--- a/atomic_defi_design/Dex/Components/GasInfoTrigger.qml
+++ b/atomic_defi_design/Dex/Components/GasInfoTrigger.qml
@@ -1,0 +1,8 @@
+import QtQuick 2.15
+
+DefaultMouseArea {
+    id: mouse_area
+    property bool no_default: true
+    anchors.fill: parent
+    onClicked: no_default ? gas_info_modal.open() : () => {}
+}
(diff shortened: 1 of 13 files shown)


cipig (approved):

atomic_defi_design/Dex/Components/GasInfoTrigger.qml
```diff
@@ -0,0 +1,8 @@
+import QtQuick 2.15
```
SylEze, line 1: No need to create a new component it's used only in one place
Moreover using directly `gas_info_modal` here leads to undefined behavior since you are not 100% sure this component will be used inside the `Dashboard` page which contains the `gas_info_modal` definition

SylEze (changes requested):

atomic_defi_design/Dex/Components/GasInfoTrigger.qml
smk762: I've created `DefaultInfoTrigger` component and replaced using `CexInfoTriger` and `GasInfoTrigger` with this.

SylEze (approved):

cipig (approved):

PR #2032, opened 2022-10-28
Debug builds

smk762:
The last PR's base branch changed to `fix/loop-crash` before merge. Current `dev` CI seems to be failing on OSX, so maybe we can merge this into `dev` before `fix/loop-crash`


smk762:
We seem to be having some issues in dev branch with CI builds, so might be worth merging this one in.

cipig (approved):

PR #2057, opened 2022-11-09
remove opt and arb

smk762:
refs:
- https://twitter.com/infura_io/status/1575138372884070400
- https://github.com/KomodoPlatform/coins/pull/496


cipig:
was the intentio to remove ARB20 completeley or only the test coins (which are affected by infura)?
because we still have ETH-ARB20, CELR-ARB20 and USDT-ARB20 in coins file
CELR-ARB20 and USDT-ARB20 are set to wallet_only because it didn't worked, so just for testing future implementations

smk762:
I'm not sure these are being used at all, and / or if the external changes are compatible. If you think they should stay and can confirm they are working, we can keep them s long as you take them for a walk every day and clean up after them.

cipig:
no, they are not working, that's why they are set to wallet_only... would be only for testing, so they work in the future... if that happens, we can also add this code back
but this PR does not actually remove them, or does it? i mean those entries are still in coins file and as such in coins_config.json

PR #2060, opened 2022-11-09
Update eth family activation

smk762:
@SylEze Seems there is `Trace/breakpoint trap (core dumped)` being triggered, and I think it is related to actions triggered by coin activation. Always happens with `[07:53:15] [debug] [qt.portfolio.model.cpp:154] [804589]: updated currency values of: TICKER` at end of logs. Happens faster when more coins enabled/enabling, and the actions triggered by the event cascade in a way that the flood of activation event triggers causes these action to be performed in quick succession for all coins multiple times in a second. Example in video:

https://user-images.githubusercontent.com/35845239/200968520-487117cb-5dcf-45e0-bd75-3b11339dd459.mp4

This looks like it might be too complex to resolve before next release (tho happy to be proven wrong), so probably best if we push this to the next iteration. The existing method can still be used with new API so it is not critical to include this yet.
There is a chance that prior crash reports may have been caused by this issue also, but these changes have made it more likely due to the increased categories of coins to enable (+1 for each eth-like protocol). Please have a look when you get a change to see if anything stands out.

cc: @tonymorony


cipig:
can we please take the contract addresses from the coins config of the respective main tokens (ETH, BNB...) instead of from the file src/core/atomicdex/constants/erc.constants.hpp ?

smk762:
Closing, stale

PR #2063, opened 2022-11-10
Standardise password length limit

smk762:
closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2054
To test:
- go thru wallet creation processes to confirm new length req message is shown.
- try to input a 256 char password into every password input field in the app, then click eyeball icon to confirm input not cut shorter
- confirm creating camouflage password still allows a single char.

### DIFF
--- a/atomic_defi_design/Dex/Components/DexAppPasswordField.qml
+++ b/atomic_defi_design/Dex/Components/DexAppPasswordField.qml
@@ -20,7 +20,7 @@ DexAppTextField
     height: 50
     width: 300
     background.radius: 25
-    max_length: General.max_pw_length
+    max_length: General.max_std_pw_length
     field.echoMode: TextField.Password
     field.font: Qt.font(
     {
--- a/atomic_defi_design/Dex/Components/DexAppPasswordField.qml
+++ b/atomic_defi_design/Dex/Components/DexAppPasswordField.qml
@@ -6,6 +6,7 @@ import Qaterial 1.0 as Qaterial
 
 import App 1.0
 import Dex.Themes 1.0 as Dex
+import "../Constants"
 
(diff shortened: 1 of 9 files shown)


SirSevenG (changes requested):
New req messages are correct ✅ 
Password fields are cut on 257th symbol, 256 fit just right ✅
Able to create 1 character camouflage pass ✅ 
Able to create 400 char camouflage pass 🛑

smk762:
The cammo password field is not governed by the same constraints, but will not work if too long (including regular passwork) for login input to accept. 
For this reason, perhaps we should limit standard password to 250 char, and limit cammo suffix to 6 char for total of 256.
@tonymorony is this reasonable? If so will implement today.

tonymorony:
> For this reason, perhaps we should limit standard password to 250 char, and limit cammo suffix to 6 char for total of 256.

@smk762 perhaps it breaks backward compatibility if user set longer camo password on previous version?

smk762:
I'm open to suggested numbers. This fix arose from an issue where user had 47 char password but privacy password input was capped at 40. 
Should I just open the limit up to 256 for both (512 total)? This would cover most users I assume.

tonymorony:
> Should I just open the limit up to 256 for both (512 total)? This would cover most users I assume.

Yep, lets do it this way to let the user decide on how to use this feature

smk762:
max input for:
- wallet creation: 256
- camo password: 256
- delete wallet: 256
- disable privacy: 256
- view seeds: 256
- Log in to wallet: 512 (to allow for standard password + camo suffix)

SirSevenG (approved):
lgtm

Canialon (approved):

PR #2070, opened 2022-11-12
ensure all protocols activation state persists after logout

smk762:
I noticed SLP tokens were not active on login. This PR makes sure they are marked in user coins file correctly to ensure this happens. Rather than setting `active` to `true` within the activation functions, this moves it to the update loop for coins returned by `get enabled coins`. This de-duplication might also potentially help with the issues mentioned in https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/2060

To test:
- for each protocol (can group EVMs into one), make sure that whatever you have enabled on logging out it auto activated when you log back in.


tonymorony (approved):
reproduced bug you've described on another branch and not able to reproduce now looks like it's fixed, thanks!

SylEze (approved):

PR #2071, opened 2022-11-13
Allow additional default coins

smk762:
To test:
- do an assets reset. 
- KMD, LTC and BTC should be enabled by default
- You should not be able to disable KMD or LTC, but can disable and BTC


atomic_defi_design/Dex/Constants/General.qml
```diff
@@ -592,7 +592,7 @@ QtObject {
         if (prevent_coin_disabling.running)
             return false
 
-        if (ticker === atomic_app_primary_coin || ticker === atomic_app_secondary_coin) return false
+        if (ticker === atomic_app_primary_coin || ticker === atomic_app_secondary_coin || ticker === atomic_app_third_coin) return false
```
tonymorony, line 595: I don't think it's a good approach to hardcode "default" coins. What if we'll want to have 10 or 20 activated on first user login?

smk762: We can make it a list. The 1st and 2nd are used to initiate base/rel so they need to be their own variable too. Avtual coins are defined in build metadata  (to allow for differences in whitelabels)

smk762: I've tweaked it so adding coins to `g_default_coins` will ensure they are auto activated on login. For demonstration purposes, https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/2071/commits/720660aad68a82bc8173c83c827ca695dffb811c includes RICK &  MORTY in this list, but will remove before final review/merge

SylEze:
Can we introduce this one in the next release please ? I'm sure it will lead to many side effects

smk762:
I can remove the "third default coin" as it is not needed if we have a "default enabed coins" list.
The issue is we need to ensure BTC is also activated automatically. Previously this was done in the local coins.json file.
Now we get the coins file from the coins repo, and before this PR, the default primary and secondary coins were marked as `"active": true` when initialising config on login.

If we have to, I can revert parts of https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/2000/files which grab the coins json from coins repo, then manually update the local on to be used on login. Hopefully there arent too many related / reliant other changes which also need to be rolled back.

I'd suggest we test this one vigilantly to confirm/deny the presence of potential issues in the meantime - if it passes muster we can continue with confidence. In the meantime I'll look at what needs to be rolled back and then we can test that to ensure it operates as expected.

tonymorony:
Also note that different white labeled wallets do have different sets of default coins (e.g. in ShibaDEX it's 20 or so), unlikely we'll create vars `atomic_app_fourth_coin` `atomic_app_fifth_coin`  `...atomic_app_n_coin`

it needs to be some list of "active by default" assets which is possible to disable

and because of the need to sync whitelabeled wallets with patch we need to do it now it seems @SylEze

smk762:
> Also note that different white labeled wallets do have different sets of default coins (e.g. in ShibaDEX it's 20 or so), unlikely we'll create vars `atomic_app_fourth_coin` `atomic_app_fifth_coin` `...atomic_app_n_coin`
> 
> it needs to be some list of "active by default" assets which is possible to disable
> 
> and because of the need to sync whitelabeled wallets with patch we need to do it now it seems @SylEze

we dont need to add vars for each additional coin. List will work, will push soon to test. list exists in this PR, but I just need to add option to disable any default except primary and secondary.

smk762:
Reopening for testing, and potential future use.

smk762:
> Can we introduce this one in the next release please ? I'm sure it will lead to many side effects

@SylEze please recheck this, I've removed "third coin" stuff and made it less obtrusive to the overall code.

SylEze (approved):
Good for me atm but please let's open an issue for next release so we make it cleaner !

SirSevenG (approved):
`+`

Canialon (approved):

PR #2072, opened 2022-11-13
Disallow disabing of parent while token active

smk762:
To Test:
- Activate a token for every EVM protocol.
- Ensure parent coin activates alongside it
- Try to disable parent coin in wallet and portfolio page while token is enabled. It should be greyed out.
- Disable all the tokens for the EMV protocol but keep parent coin enabled.
- Try to disable parent coin in wallet and portfolio page while no token is enabled. It should possible.

### DIFF
--- a/atomic_defi_design/Dex/Components/CoinMenu.qml
+++ b/atomic_defi_design/Dex/Components/CoinMenu.qml
@@ -9,6 +9,7 @@ Menu {
     Universal.accent: Style.colorQtThemeAccent
     Universal.foreground: Style.colorQtThemeForeground
     Universal.background: Style.colorQtThemeBackground
+    property bool can_disable;
 
     // Ugly but required hack for automatic menu width, otherwise long texts are being cut
     width: {
@@ -28,7 +29,7 @@ Menu {
         id: disable_action
         text: qsTr("Disable %1", "TICKER").arg(ticker)
         onTriggered: API.app.disable_coins([ticker])
-        enabled: General.canDisable(ticker)
+        enabled: can_disable
     }
 
     MenuItem {
--- a/atomic_defi_design/Dex/Constants/General.qml
+++ b/atomic_defi_design/Dex/Constants/General.qml
@@ -579,22 +579,35 @@ QtObject {
         return exists(v) && v !== ""
     }
 
-    function isParentCoinNeeded(ticker, type) {
-        for(const c of API.app.portfolio_pg.get_all_enabled_coins())
-            if(c.type === type && c.ticker !== ticker) return true
-
+    function isParentCoinNeeded(ticker, coin_type)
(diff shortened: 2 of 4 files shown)


cipig (dismissed):

Canialon (changes requested):
>Try to disable parent coin in wallet and portfolio page while token is enabled. It should be greyed out.

I guess it makes sense to add some info about why the user cant deactivates the parent coin. Now it is just unclear. need some info wny user cant deactivate it
image

_______

Also, some parents cant be deactivated even if there are no active tokens. At least  bnb, avax and ftm(upd: can deactivate ftm after double token activation and diactivation)
image

Canialon:
>Disable all the tokens for the EMV protocol but keep parent coin enabled.

Parent coin deactivates with all tokens 


https://user-images.githubusercontent.com/77272165/201622512-ba68db86-7ba2-4739-9e6c-4a290e23e6ba.mov

smk762:
> > Disable all the tokens for the EMV protocol but keep parent coin enabled.
> 
> Parent coin deactivates with all tokens
> 
>  Screen.Recording.2022-11-14.at.09.20.15.mov

This as the previous result also perhaps we can improve the message to say "Disable BNB and ALL BEP20 assets". BNB is tagged as "type:BEP-20" so internally it is considered a BEP20 asset. I'm not sure if we want the EVM parent to not disable in this case.

smk762:
> I guess it makes sense to add some info about why the user cant deactivates the parent coin. Now it is just unclear. need some info why user cant deactivate it 

Lets add this as an issue for later. There is some existing code which would show a pop up saying "cant deativate default coin" which I dont think is in use at the moment, but that is a bad ux flow. Maybe we ca have a tooltip in the context menu :thinking: 

> Also, some parents cant be deactivated even if there are no active tokens. At least bnb, avax and ftm(upd: can deactivate ftm after double token activation and deactivation) 

I'll check to replicate here - there may be a slight delay before app knows tokens are inactive so if its just like that for a few seconds after, that might explain it.

SirSevenG:
> Parent coin deactivates with all tokens

It makes sense to me at least. We my change deactivation phrase tho, to make it more clear for enduser.

Canialon:
okay it make sense
but this moment is still here, cant deactivate avax and bnb:
image

smk762:
Thanks for feedback @Canialon - the BNB AVX FTM problems should now be fixed

smk762:
Good question. I think we can leave that for next iteration, because at the moment is has been a bit hybrid while guis have had slightly different coins file requirements. for 0.5.8 we can resolve that more accurately in the back end.

Canialon (approved):

SirSevenG (approved):

PR #2097, opened 2022-11-17
remove zhtlc_enter_enabling and zhtlc_leave_enabling

smk762:
Potential solution to https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2094

These events don't add any real value beyond notifications, and was poorly implemented as it was not linked to a ticker so removing it seems like a good idea. 
After a couple of test runs including these changes, it seems to overcome https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2094 though further stress testing to confirm should be done.

To confiirm,  ensure no crash events caused by:
- enabling or disabling bulk coins alongside in-progress ZHTLC activation
- logging out during in-progress ZHTLC activattion, then back in to same or different wallet without closing app.

### DIFF
--- a/src/core/atomicdex/events/events.hpp
+++ b/src/core/atomicdex/events/events.hpp
@@ -30,8 +30,6 @@ namespace atomic_dex
     using gui_leave_trading     = entt::tag<"gui_leave_trading"_hs>;
     using mm2_initialized       = entt::tag<"mm2_running_and_enabling"_hs>;
     using default_coins_enabled = entt::tag<"default_coins_enabled"_hs>;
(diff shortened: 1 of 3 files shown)


smk762:
upon further it looks like the fix is not this simple - but these are worth removing anyway so will continue to attempt a fix in this branch

smk762:
I can replicate a crash without zhtlc by spamming enable / disable rapidly in sequence, but have been unable to replicate the original assumed ZHTLC related crash encountered under non-spam conditions.

This is very confusing.

SylEze (approved):

PR #2101, opened 2022-11-18
close -> cancel + translated

smk762:
closes: https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2093


Canialon:
I believe it makes sense to do the same for 
- the "enable assets" window
image

- send window
image

SylEze (approved):

PR #2124, opened 2022-11-28
Changes asks list order

### LINKED ISSUE #2116: [BUG]: Orderbook entries are reversed
Right now top (red) part of the orderbook is suboptimal - best orders are on the bottom and worst are on top what makes interface inconsistent, and in case of orderbook update user have to scroll from top to the bottom to see best order / loses focus from best order what a) just not good in terms of UX b) can cause not best order matching

We should reverse it's content and show best orders on top (same as in green part) and ensure that best order is always on top on orderbook refresh without the need to scroll anything 

https://user-images.githubusercontent.com/24797699/204137086-4cc37b47-3553-469c-812f-f3efc6311651.mov

SylEze:
Closes #2116

### DIFF
--- a/atomic_defi_design/Dex/Exchange/Trade/OrderBook/List.qml
+++ b/atomic_defi_design/Dex/Exchange/Trade/OrderBook/List.qml
@@ -13,8 +13,6 @@ Item
 
     property bool isAsk
     property bool isVertical: false
-    property alias view: orderbook_list
-
     width: parent.width
     height: parent.height
 
--- a/atomic_defi_design/Dex/Exchange/Trade/OrderBook/List.qml
+++ b/atomic_defi_design/Dex/Exchange/Trade/OrderBook/List.qml
@@ -13,6 +13,8 @@ Item
 
     property bool isAsk
     property bool isVertical: false
+    property alias view: orderbook_list
+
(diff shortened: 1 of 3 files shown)


SirSevenG (dismissed):
lgtm

tonymorony (changes requested):
asks list jumping somewhere to the middle when I select bid and I have to scroll back:

https://user-images.githubusercontent.com/24797699/204336345-c1ff3283-df5b-4e7f-bba1-39229f510e92.mov

smk762:
> asks list jumping somewhere to the middle when I select bid and I have to scroll back:

replicated here

smk762:
After switching tickers seems the scroll is not where it should be.

https://user-images.githubusercontent.com/35845239/204824319-880e623d-c0ea-4eaa-b591-2660dba2030b.mp4

SylEze:
Let's discuss on how the orderbook can be improved in another issue, this PR just aims to change the orderbook sorting order and has no effet on what has been mentioned

cipig:
i disagree, see https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2116#issuecomment-1333126923
the sorting was correct

PR #2134, opened 2022-11-30
Fix trade confirm modal ticker icons wrong

### LINKED ISSUE #2105: [BUG] Trade confirmation tickers not same as selected in order
image
Parts of it say KMD/Komodo and none of it should when making a TKL/ARRR trade

SylEze:
Closes #2105

### DIFF
--- a/atomic_defi_design/Dex/Components/PairItemBadge.qml
+++ b/atomic_defi_design/Dex/Components/PairItemBadge.qml
@@ -3,10 +3,11 @@ import QtQuick.Layouts 1.15
 
 import bignumberjs 1.0
 
+import "../Constants"
+
 DefaultRectangle
 {
     id: root
-    property alias  source: icon.source
     property alias  ticker: ticker.text
     property alias  fullname: fullname.text
     property string amount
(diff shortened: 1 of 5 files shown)


tonymorony:
accordingly to the reports the problem is not just in icon but actually wrong tickers passing (https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2105#issuecomment-1327406486 there is an marketmaker error that Base/Rel tickers can't be the same in this case)

Was you able to reproduce the issue on your side @SylEze ?

SylEze:
> accordingly to the reports the problem is not just in icon but actually wrong tickers passing ([#2105 (comment)](https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2105#issuecomment-1327406486) there is an marketmaker error that Base/Rel tickers can't be the same in this case)
> 
> Was you able to reproduce the issue on your side @SylEze ?

The only issue was the icons on my side
If you look at the text below, the tickers displayed are the good ones and I was able to make a swap when the icons were wrong

smk762 (approved):
I have been unable to replicate the initial issue

SirSevenG (approved):
Was not able to reproduce intial issue

PR #2157, opened 2022-12-05
Update style.qml

estar-app:
Update style.qml file for both NATURE and SIXCOIN.


smk762 (changes requested):
missing commas - you need to add these at the end of the lines for runes and six or else it wont parse correctly.

smk762 (approved):

PR #2178, opened 2022-12-11
limit warning text values to 8dp

smk762:
closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2177

To test:
- simulate case where low balance error happens in pro view (you can do this by broadcasting a withdrawal in CLI using the app RPC credentials, then trying to do a swap soon after)


cipig (approved):
image

SylEze (approved):

PR #2200, opened 2023-01-15
Added vprm.png

VaporumCoin:
Added Vaporum Coin VPRM logo


cipig:
there is no need to add the icons to Desktop repo any more, they are now fetched from coins repo automatically on each build... and it works fine
image

PR #2201, opened 2023-01-21
no console logs for release builds

smk762:
Closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2171
To test:
- Launch on linux from file manager.
- In console, do `tail -f /var/log/syslog`.
- You should not see any Atomicdex related logs in syslog.
- Confirm launch on Mac and Windows works as expected.

cc: @himu007

### DIFF
--- a/src/app/main.prerequisites.hpp
+++ b/src/app/main.prerequisites.hpp
@@ -1,5 +1,5 @@
 /******************************************************************************
- * Copyright © 2013-2021 The Komodo Platform Developers.                      *
+ * Copyright © 2013-2023 The Komodo Platform Developers.                      *
  *                                                                            *
  * See the AUTHORS, DEVELOPER-AGREEMENT and LICENSE files at                  *
  * the top-level directory of this distribution for the individual copyright  *
@@ -180,15 +180,15 @@ static void init_logging()
     auto rotating_sink = std::make_shared<spdlog::sinks::rotating_file_sink_mt>(path.string(), spdlog_max_file_size, spdlog_max_file_rotation);
 #endif
 
+#if defined(DEBUG) || defined(_WIN32) || defined(WIN32)
     std::vector<spdlog::sink_ptr> sinks{stdout_sink, rotating_sink};
+#else
+    std::vector<spdlog::sink_ptr> sinks{rotating_sink};
+#endif
     auto logger = std::make_shared<spdlog::async_logger>("log_mt", sinks.begin(), sinks.end(), tp, spdlog::async_overflow_policy::block);
     spdlog::register_logger(logger);
     spdlog::set_default_logger(logger);
-#ifdef DEBUG
     spdlog::set_level(spdlog::level::trace);
-#else
-    spdlog::set_level(spdlog::level::info);
-#endif
     spdlog::set_pattern("[%T] [%^%l%$] [%s:%#] [%t]: %v");
 }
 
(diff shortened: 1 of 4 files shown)


tonymorony:
I believe the problem mentioned in issue was not the logs itself but the fact that logs writing into `syslog` file

what is the QT feature introduced in 0.5.6

https://www.ics.com/blog/whats-new-qt-560-logging-syslog-and-journald

```
To enable the syslog or journald logging support in Qt 5.6.0 or later, you need to configure it into Qt. They default to being disabled, so you need to explicitly use the -journald or -syslog options when running the configure script. If successfully enabled, you should see the relevant output of the configure script indicate it as being enabled, as shown below:

  Logging backends:
    journald ............... yes
    syslog   ............... yes
```

so just logging backend removal is needed

why do we need temp.txt log file?

smk762:
I saw that article when investigating the problem.

>  you need to configure it into Qt. They default to being disabled
App is not configured to write to syslog. 

> you need to explicitly use the -journald or -syslog options when running the configure script.
The strings `syslog` & `journald` do not exist anywhere in the source code, workflows or build scripts.

> why do we need temp.txt log file?
This is a way to get the response returned from https://github.com/KomodoPlatform/atomicDEX-Desktop/blob/dev/src/core/atomicdex/utilities/kill.cpp#L29 so that `mm2: no process found` wont show up in console / syslog when app tries to kill existing instances of `mm2`. 

More context: https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2171#issuecomment-1399162497

cipig (approved):
this disables console log completely
if that is the intention, then it works
can't tell anything about syslog though, it didn't had that problem on linux myself

smk762:
> can't tell anything about syslog though, it didn't had that problem on linux myself

Did you launch from file manager? If so do you know which one?

cipig:
> Did you launch from file manager? If so do you know which one?

Oh... no, i didn't... just unpacked and started `AntaraAtomicDexAppDir/usr/bin/atomicdex-desktop` from console

himu007:
@smk762 do you have a build file that i can test?

cipig:
Ok, i unzipped on CLI and started with FileManager (which doesn't make that much sense, since it's more work to switch to file manager and click me through till i get to the place where the binary is located... but ok, just to test)
seeing this in syslog
```
Jan 23 17:50:59 Flow-PC systemd[1202]: Started Dolphin - File Manager.
Jan 23 17:51:23 Flow-PC systemd[1202]: Started /home/florian/Tools/AntaraAtomicDexAppDir/usr/bin/atomicdex-desktop.
```
but that's it, nothing more in syslog
btw, you need to do `sudo tail -f /var/log/syslog`, because syslog can't be read by random users, only root can do that

himu007:
@cipig 
> btw, you need to do sudo tail -f /var/log/syslog, because syslog can't be read by random users, only root can do that
my Ubuntu 20.04 Desktop is installed with one 1 user (non-root) account and i can tail the syslog file without `sudo`. It's just the default installation using the desktop ISO file from Ubuntu website without any extra setup/steps.

though it is not possible in my server running Debian, where user setup is same. i am guessing it varies depending on OS/flavor.

himu007:
@smk762 it appears the issue i was seeing is fixed with builds from this PR
only entry in syslog i found after i started the AppImage file is:
```
Jan 23 18:11:48 x org.gnome.Nautilus[209740]: [2023-01-23 18:11:48.306] [info] [global.utilities.cpp:65] creating directory /home/user/.atomic_qt/logs
```
i no longer see the full logs being shown in syslog file.

SirSevenG (approved):
Haven't seen app logs in syslog here

smk762:
@tonymorony please merge if you are happy with this PR in its current state.

tonymorony (approved):

PR #2213, opened 2023-03-02
suppress matic balance notifications

smk762:
There is currently an issue where users will get unexpected notifications about matic balance due to inconsistent upstream data. This may be due to hitting api rate limits, though I've not been able to investigate thorougly yet.
In the meantime, we can suppress notifications related to MATIC balance changes to keep UX intact.


SirSevenG:
> i fear, the issue is from specific RPC endpoints that have a problem

There were long maintenance for matic public resources a while ago. Data from endpoints were inconsistent and pretty random.
There's a chance UI notifications are related to that.

SirSevenG:
> we can use our quicknode endpoints instead of public ones lets say, I think it should solve the problem

Should work, imo.

smk762:
Closing in favour of https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/2213
This does leave us with a "single point of failure" though. We should add backups for each of our quicknodes.

PR #2227, opened 2023-03-28
Update version to 0.5.7.2 to add VOTE2023 and other new coins

smk762:
to test:
- check app launches on all OS.
- confirm VOTE2023 is a default, wallet only coin.


Canialon (approved):
lgtm:) thanks

cipig (approved):

endrilickollari (approved):
VOTE2023 coin is a default, wallet only. Can send & receive but can't swap.

image

PR #2228, opened 2023-03-30
remove duplicate coincolor code

cipig:
the same list is already in atomic_defi_design/Dex/Constants/Style.qml


smk762:
To test:
- launch app
- confirm coin text / tags displays correctly in all pages (enable, wallet, transaction confirmation, transaction history, portfolio, dex simple, dex pro, addressbook, swap confirmation, swap status, swap history etc).

Canialon:
@cipig can you please marge `dev` into this branch to trigger CI builds 🙏

cipig:
> @cipig can you please marge `dev` into this branch to trigger CI builds pray

done... but MacOS will not build, since i have done this PR from my own repo... forgot that MacOS is not built when i do that, sorry

cipig:
oh, guess that wasn't the best idea... i merged dev into it, but this PR was moved to 0.5.8 branch
looks weird atm... maybe merging dev to 0.5.8 will clear it up again

endrilickollari (commented):
Is it okay that `Arbitrum` has same color as `ERC-20`: 

image

image

endrilickollari:
> Is it okay that `Arbitrum` has same color as `ERC-20`:
> 
> image
> 
> image

Text color is okay: 
BEP-20 != others: 

image

cipig:
the colors of the chains were not modified in this PR, i just removed the duplicate definition of coin colors
we should actually remove that thing completely and replace it with chain specific colors, so we don't have to define a color for each coin in part... if we want to keep colors for each coin, we should move the config into coins repo
but i think use chain colors will be better... would also give us the option to remove the chain suffix from tickers in some places if the different chains are differentiated by different colors

endrilickollari:
> the colors of the chains were not modified in this PR, i just removed the duplicate definition of coin colors we should actually remove that thing completely and replace it with chain specific colors, so we don't have to define a color for each coin in part... if we want to keep colors for each coin, we should move the config into coins repo but i think use chain colors will be better... would also give us the option to remove the chain suffix from tickers in some places if the different chains are differentiated by different colors

Beside that, everything looks okay to me.

smk762 (approved):

PR #2229, opened 2023-03-31
remove band oracle code

cipig:
closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2031


smk762:
For future reference - no need to update translation files. These are generated at build time.

PR #2230, opened 2023-03-31
Update API

smk762:
This branch is for updating 0.5.7.x to 0.5.8 once the required methods and changes have been implemented for compatibility with the v1.0.x-beta API. 

Closes:
- https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2109
- https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2046

Done:
- ZHTLC method names and request/response structures updated
- Resolved a P0 orderbook/bestorders bug which caused crashes due to duplicated uuids in response for segwit coins.
- Added Tendermint activation
- Added memo sending for ZHTLC and Tendermint coins

To Test:
- Activate ZOMBIE, use faucet, send funds, do a swap
- Activate IRIS, use faucet, send funds with memo (confirm memo on explorer), do a swap
- Activate ARRR and other tendermint assets
- Try to crash the pro view orderbook (click on things like a mad man)

You likely will need to do a wallet reset to see IRIS and other new coins.


src/core/atomicdex/models/qt.orderbook.model.cpp
```diff
@@ -374,13 +375,22 @@ namespace atomic_dex
         m_orders_id_registry.clear();
         for (auto&& order: m_model_data)
         {
+            // Maybe adding a suffix for segwit entries could avoid this?
+            // Working so far, but needs more testing to see if there are ripple effects due to the uuid suffix.
+            // Early tests confirm order selection and placing do not appear to be negatively affected.
```
cipig, line 380: the effect i observed from this -segwit change is that you first see the -segwit entries in best-orders for couple seconds, then they are removed and only the non-segwit are left
i think it's good to only show one entry per coin, the duplicated entries with segwit were a bit annoying
problem is though that now people are not aware that they could also trade with segwit, saving some fees (though question is how much fees they actually save... i guess it makes a difference when using many inputs, but with a single input it will likely not make much of a difference)
this only applies to BTC, the other UTXO coins have such low fees that segwit is not needed for this
so i was thinking about this:
- make BTC-segwit default coin instead of BTC: https://github.com/KomodoPlatform/atomicDEX-Desktop/blob/dev/src/core/atomicdex/constants/dex.constants.hpp#L14 ... or add it next to BTC
- invert this filter for BTC and show the BTC-segwit entries instead of BTC in best-orders
what do you think about this?

smk762: both the segwit and non segwit entries are exactly the same apart from ticker. The filtering is implied, I assume the orderbook model removes the nonstandard uuids. It could be a bit of work to get around this, so please add an issue for it. 
re: Default coins, I think having both would be good as it also lets the user know there is both available.

cipig:
image
shows up when i click on "start swap" to buy IRIS with RVN

the error from mm2 is:
`[17:31:45] [info] [qt.trading.page.cpp:1207] [4182732]: [determine_fees] trade_preimage answer received: [{"error":"rpc:211] dispatcher_legacy:141] dispatcher_legacy:158] taker_swap:2358] tendermint_coin:1488] tendermint_coin:958] Transport error: Could not read gas_info. Error code: 6 Message: rpc error: code = Unknown desc = failed to execute message; message index: 0: 2000000uiris is smaller than 262520607uiris: insufficient funds [cosmos/cosmos-sdk@v0.46.5/x/bank/keeper/send.go:191] With gas wanted: '18446744073709551615' and gas used: '59863' : unknown request"}]`

when i do `get_trade_fee` on IRIS i get `"error": "rpc:211] dispatcher_legacy:141] lp_coins:3334] Not implemented"` in CLI

endrilickollari (approved):
1. Successfully activated `ZOMBIE` asset. 
    
image

Did send funds, also swap coin: 

image

image

2. Successfully activated `ARRR `and Tendermint assets: 

    
image


3. Send fund with `IRIS`, swap, use memo when sending funds: 

    
image


image


image


image



OS: Windows 10

Canialon:
hey @smk762 I cannot see the transaction history for TENDERMINT
image

image

tested on MacOS using CI build

src/core/atomicdex/models/qt.orderbook.model.cpp
smk762: resolved in https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/2280

endrilickollari:
1. Successfully activated `ZOMBIE`: 
    
    
image

Sent was successful: 

image

Swap was successful: 

image

2. `IRIS` successfully activated:

image

Transactions history not visible. 

Memo visible in explorer view: 
image

image

image

Swap was also successful: 
image


3.. Tendermint coins successfully activated: 
image
while for two coins there is this error: 

```
11 10:05:23, mm2_main::mm2::rpc::dispatcher:113] ERROR RPC error response: platform_coin_with_tokens:314] tendermint_with_assets_activation:191] Error on platform coin OSMO creation: tendermint_coin:2231] tendermint_coin:440] All the current rpc nodes are unavailable.
11 10:05:23, mm2_main::mm2::rpc::dispatcher:113] ERROR RPC error response: platform_coin_with_tokens:314] tendermint_with_assets_activation:191] Error on platform coin ATOM creation: tendermint_coin:2231] tendermint_coin:440] All the current rpc nodes are unavailable.
```

cipig (approved):

PR #2235, opened 2023-04-03
V2 erc activation

smk762:
- renames some method functions / files to indicate v2 rpc
- adds functions for `enable_erc20` and `enable_eth_with_tokens`

Unfortunately using the new methods is causing a `Trace/breakpoint trap (core dumped)` so though they are present in the code, they are not in use at the moment. More testing needs to be done to isolate and resolve the cause.


cipig:
this PR should be prioritized
atm no PLG20 token is enabled at start, i get error about too many requests from the public RPC nodes
v2 erc activation with param `\"get_balances\":false` would fix the problem, it works fine on all my 6 mm nodes

src/core/atomicdex/services/mm2/mm2.service.cpp
```diff
@@ -280,21 +281,23 @@ namespace atomic_dex
...
-        const auto now    = std::chrono::high_resolution_clock::now();
-        const auto s      = std::chrono::duration_cast<std::chrono::seconds>(now - m_orderbook_clock);
-        const auto s_info = std::chrono::duration_cast<std::chrono::seconds>(now - m_info_clock);
+        const auto now                 = std::chrono::high_resolution_clock::now();
+        const auto s_orderbook         = std::chrono::duration_cast<std::chrono::seconds>(now - m_orderbook_clock);
+        const auto s_coins_status      = std::chrono::duration_cast<std::chrono::seconds>(now - m_coin_status_clock);
+        const auto s_info              = std::chrono::duration_cast<std::chrono::seconds>(now - m_info_clock);
 
-        if (s >= 5s)
+        if (s_coins_status >= 15s)
```
smk762, line 289: leaving this out of next integration attempt as a potential source of crashing.

PR #2237, opened 2023-04-03
update vcpkg action version

smk762:
To test:
- please download binaries from https://github.com/KomodoPlatform/atomicDEX-Desktop/actions/runs/4596910754
- Launch the portable / installer versions for each OS you can (note which OS you confirmed in review comments).
- Basic smoke test of app (login, do swap)

### DIFF
--- a/.github/workflows/atomicdex-desktop-ci.yml
+++ b/.github/workflows/atomicdex-desktop-ci.yml
@@ -22,6 +22,7 @@ env:
   RUNVCPKG_VCPKG_ROOT: ${{ github.workspace }}\atomicDEX-Desktop\ci_tools_atomic_dex\vcpkg-repo
   VCPKG_ROOT: ${{ github.workspace }}\atomicDEX-Desktop\ci_tools_atomic_dex\vcpkg-repo
 
+
 jobs:
 
   ci-flow:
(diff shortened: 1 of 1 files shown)


cipig (approved):
works fine under debian 11

endrilickollari (dismissed):
Looks very good on windows 10.

.github/workflows/atomicdex-desktop-ci.yml
```diff
@@ -181,11 +193,11 @@ jobs:
           echo "CC=clang-12" >> $GITHUB_ENV
 
       - name: vcpkg deps (All)
-        uses: lukka/run-vcpkg@v10
+        uses: lukka/run-vcpkg@v11
         with:
           vcpkgDirectory: '${{ github.workspace }}/ci_tools_atomic_dex/vcpkg-repo'
-          appendedCacheKey: ${{ hashFiles('vcpkg.json') }}
           vcpkgJsonGlob: 'vcpkg.json'
+          appendedCacheKey: vcpkginstall
```
cipig, line 200: saw this just now:
```
Warning: Unexpected input(s) 'appendedCacheKey', valid inputs are ['vcpkgDirectory', 'runVcpkgInstall', 'vcpkgGitCommitId', 'vcpkgGitURL', 'doNotUpdateVcpkg', 'doNotCache', 'vcpkgJsonGlob', 'vcpkgJsonIgnores', 'binaryCachePath', 'runVcpkgFormatString', 'useShell', 'logCollectionRegExps']
```
i have no idea what it does but seems like `runVcpkgInstall` would be a "correct" value

Canialon (dismissed):
Looks good on MacOS 13 and Windows 11

.github/workflows/atomicdex-desktop-ci.yml
smk762: the way caching works changed, so the old key is obsolete. `runVcpkgInstall` should remain false (as the default)

PR #2239, opened 2023-04-04
remove nomics and replace chart

smk762:
Closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2238

To test: 
- check dex pro view chart. resize / maximise window. change pair. there should be no flicker or overflow.

We dont have i or this widget yet, though it uses tickers by default so it the ticker is supported a chart will display.
This has some known issues:
- The ticker may link to the wrong coin
image
- the ticker may not be supported
image

Until we can compile a list of correct/working tickers, I'll add an exclusion list for coins known to be no good. Please list any you encounter in this PR.

### DIFF
--- a/atomic_defi_design/Dex/Constants/General.qml
+++ b/atomic_defi_design/Dex/Constants/General.qml
@@ -37,39 +37,14 @@ QtObject {
         }
     }
 
-
-    function getChartExcludeList(ticker)
-    {
-        let coin_info = API.app.portfolio_pg.global_cfg_mdl.get_coin_info(ticker)
-        let exclude_list = ["RICK", "MORTY", "ZOMBIE"]
-        if (ticker.search("TEST") > -1 || ticker.search("t") == 0 || coin_info.is_testnet)
-        {
-            exclude_list.push(ticker)
-        }
-        return exclude_list
-    }
-
-
     function getChartTicker(ticker)
     {
-        ticker = coinWithoutSuffix(ticker)
-        let exclude_list = getChartExcludeList(ticker)
-        let redefine_list = {}
-        if (exclude_list.includes(ticker))
-        {
-            return ""
-        }
-        if (redefine_list.hasOwnProperty(ticker))
-        {
-            return redefine_list[ticker]
-        }
-        return ticker        
+        let coin_info = API.app.portfolio_pg.global_cfg_mdl.get_coin_info(ticker)
+        return coin_info.livecoinwatch_id
     }
-
 
     function coinWithoutSuffix(ticker)
     {
-        console.log("coinWithoutSuffix: " + ticker)
         if (ticker.search("-") > -1)
         {
             return ticker.split("-")[0]
@@ -78,6 +53,12 @@ QtObject {
         {
             return ticker
         }
+    }
+
+    function is_testcoin(ticker)
+    {
+        let coin_info = API.app.portfolio_pg.global_cfg_mdl.get_coin_info(ticker)
+        return coin_info.is_testnet
     }
 
     function coinName(ticker) {
--- a/atomic_defi_design/Dex/Constants/General.qml
+++ b/atomic_defi_design/Dex/Constants/General.qml
@@ -37,6 +37,36 @@ QtObject {
         }
     }
 
+
+    function getChartExcludeList(ticker)
+    {
+        let coin_info = API.app.portfolio_pg.global_cfg_mdl.get_coin_info(ticker)
+        let exclude_list = ["RICK", "MORTY", "ZOMBIE"]
+        if (ticker.search("TEST") > -1 || ticker.search("t") == 0 || coin_info.is_testnet)
+        {
+            exclude_list.push(ticker)
+        }
+        return exclude_list
+    }
+
+
+    function getChartTicker(ticker)
+    {
+        ticker = coinWithoutSuffix(ticker)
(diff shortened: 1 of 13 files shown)


cipig:
we could add a new file to https://github.com/KomodoPlatform/coins/tree/master/api_ids for livecoinwatch
if there is an id for the coin, it takes that (useful for duplicate tickers like TRC for https://www.livecoinwatch.com/price/Terracoin-TRC and https://www.livecoinwatch.com/price/TRACERToken-_TRC... for the other one they use `_TRC` as ticker, which we can define like that in the file)
and the cases where the ticker is there but points to a wrong coin, we can use "" in the new api_ids file, to tell ADEX to not load a chart at all... https://www.livecoinwatch.com/price/McLarenF1FanToken-MCL would be such a candidate... it uses MCL, but is not Marmarachain

can we also please use 7d charts, instead of 1d? just set `lcw-period="w"`

smk762:
> can we also please use 7d charts, instead of 1d? just set `lcw-period="w"`

Done. I've not implemented an exclusion list yet, but will populate it from review feedback.

cipig:
looks good, thanks
we need to remove the chain from the ticker though, eg `-BEP20` and also `-segwit` because:
image
image

cipig (approved):
works fine now, thanks
this error shows up when i switch pairs, but looks like it has no negative effect, and idk if it's even related
`[12:42:53] [error] [main.prerequisites.hpp:96] [304992]: Error: <svg> attribute viewBox: Unexpected end of attribute. Expected number, "".`

Canialon (commented):
- looks like it can be a scrollbar if it is no chart, i guess it is not very good, will be nice to not cause it.
image

- in pair of fiat and not fiat coin: with some test coins, it is no data but with (screenshot 1) but with some test coins the chart displays with $ (screenshot 2)
image
image


-sometimes size broken 
image


- the resizing looks weird and slow, also can just freeze with "loading data"

https://user-images.githubusercontent.com/77272165/230104854-8b4db08b-a7dd-4d8c-85ef-4c4efa982d73.mov


https://user-images.githubusercontent.com/77272165/230105951-fd5de32c-c13c-4143-ab7c-29f0960d0ad5.mov

UPD: tested on MacOS

cipig:
strange, i don't have those problems with the sidebar, the chart is not reloaded... the only thing is that everything is moved to the right, even if it doesn't fit into the window, like this:
image

when i resize, the loading is indeed somewhat slow, but i don't care, i don't resize all the time
btw, saving the window size would be nice, then i wouldn't have to resize at all, see https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2207

endrilickollari:
Sidebar loading happening in Windows 10 also, sometimes it loads faster sometimes it gets stuck also during resizing it does the same thing. 

Video: 


https://user-images.githubusercontent.com/32954585/230185992-d264056e-9b7a-44fd-8bd6-6326bc2983e5.mp4

Also when trying to trade test coins message appears: 

image


OS: Windows 10 
Resolution: 1920 x 1080

endrilickollari (commented):
Sidebar loading happening in Windows 10 also, sometimes it loads faster sometimes it gets stuck also during resizing it does the same thing.

Video:


https://user-images.githubusercontent.com/32954585/230187321-d0fd0972-3917-410a-b8cb-53d806b81153.mp4



Also when trying to trade test coins message appears:

image


OS: Windows 10
Resolution: 1920 x 1080

smk762:
> works fine now, thanks this error shows up when i switch pairs, but looks like it has no negative effect, and idk if it's even related `[12:42:53] [error] [main.prerequisites.hpp:96] [304992]: Error: <svg> attribute viewBox: Unexpected end of attribute. Expected number, "".`

This error is thrown from the widget's own script. It also comes up in the console on https://www.livecoinwatch.com/widgets

smk762:
> -sometimes size broken image
> 
> * the resizing looks weird and slow, also can just freeze with "loading data"
> 
> UPD: tested on MacOS

I've now set it to use a fixed width, which should resolve the resizing issues.
Testcoins should also now be filtered out (if either ticker is a testcoin, there will be no chart)

smk762:
> Also when trying to trade test coins message appears:

This is expected, though I might tweak the wording, because `yet` implies there will be a chart in future (but for test coins this is unlikely)

smk762:
Ready for re-review. **You will have to reset assets config for the latest changes in coins repo (https://github.com/KomodoPlatform/coins/pull/689) to apply** 

https://user-images.githubusercontent.com/35845239/230337523-8c63cd81-9e55-41fb-a457-1de5bf6ffcc6.mp4

smk762:
It seems for some minor coins which are supported by livecoinwatch may not be supported as a pair. I can't think of a method to filter these out without using a pair exclusion list.

https://user-images.githubusercontent.com/35845239/230338626-9a359d94-dd1d-4208-a2f7-38fe9302f09b.mp4

As it is only affecting less common pairs, I'll make this an issue for future (so not a blocker for this PR)

cipig (approved):

endrilickollari (commented):
Chart looks very good now

video: 


https://user-images.githubusercontent.com/32954585/230459312-6035d9f9-7786-43ec-afe3-e78b911a0820.mp4


OS: Windows 10

Canialon (approved):
looks good now:D

PR #2242, opened 2023-04-07
Fix combos overflow

smk762:
Before:
image
image
image
image

After:
image
image
image
image


cipig:
the names of most of the tokens don't fit in the drop down
image
can we make that a bit wider? though it will not help that much, i guess only removing the suffix from the tickers would really help here... unsure what to do here

on the swap detail page, the name is basically completely gone:
image
i guess we can make this wider without problems

cipig:
thanks, the coin names are now much better visible
can you please change the text color on charts? on lite mode it's not visible, because it's also white, like the background
image

Canialon:
@smk762 Does chart looks okay on your side? On my side it is like this for all coins 🤔 
image

smk762:
Sorry I should have mentioned, you'll need to do an assets reset

cipig:
this https://github.com/KomodoPlatform/coins/pull/692 helps too with the charts, but needs to be merged

Canialon:
❔  the names still don't fit in the drop down
image

but looks like it is much more space for names on details page:D
image

❗️at thr screenshot above the BNBT name is invisible 
❗️same for Tusd in the dropdown:(
image

cipig:
> the names still don't fit in the drop down

that is mainly related to the overall available width, it is 1280 atm (we have a min resolution set to 1280 x 800)
see also https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2243
i changed that in my repo to 1366 x 768 and made the chart column wider, so those dropdowns are wider too
it's not a big change, so if desirable, we could include that in the 0.5.7.2 release
the 1366 x 768 would actually cover more possible resolutions as of https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide, but it reduces the height from 800 to 768 (already wondered that nobody complained about the 800 because on 1024 x 768, the window should be cut off)

smk762:
We should create a separate issue for that extended ticker overflow. I've made those combos as wide as possible for now.
Thanks for the bad contrast coin images, I'll fix those now.

smk762:
This PR is ready for final review. If any issues found which are not critical/blocking, please make an issue for it.

cipig (approved):
closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2211

Canialon (approved):
looks good so far

PR #2245, opened 2023-04-10
Lite theme contrast fixes

smk762:
Fixes some bad contrast coloring in lite theme for buttons and "add contact" pop up.
Also removes an obsolete/duplicate component.
To test:
- Use lite theme
- Check the pop up in addressbook when you press "new contact" button. The background of the pop up should be different color to the input field.
- Check the "refresh" button in the "Claim KMD rewards modal".  It should be clearly visible and not the same color as the background.

### DIFF
--- a/assets/themes/Default - Light/colors.json
+++ b/assets/themes/Default - Light/colors.json
@@ -2,13 +2,14 @@
     "accentColor": "#F0F2FFFF",
     "foregroundColor": "#456078FF",
     "backgroundColor": "#FCFCFCFF",
-    "backgroundColorDeep": "#FCFCFCFF",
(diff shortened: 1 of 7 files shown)


cipig:
is the default coin color on lite mode white? maybe change that to default to something else... or is that intentional?
image
image
image

smk762:
> is the default coin color on lite mode white? maybe change that to default to something else... or is that intentional? 

This has been fixed in https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/2242 (https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/2242/commits/5e1e85e467f27eeeb65296ee35151d67468b6e04)

cipig (approved):

PR #2247, opened 2023-04-12
Set BTC-segwit to wallet only

smk762:
Due to variation in the version of API used between desktop and other GUIs / LPs, there have been cases of failed swaps involving `BTC-segwit` between desktop and other users. This will be resolved once we update to the latest API in 0.5.8, but until then we can make `BTC-Segwit` wallet only to avoid failed swaps.

To test:
- Login and activate `BTC-segwit`
- You should be unable to select `BTC-segwit` for trading


cipig (approved):

Canialon (approved):
looks good:)
image

PR #2262, opened 2023-04-13
handle button where no notifications

smk762:
Closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2258

To test: View your notifications when the list is empty. The button should have different text as below:

https://user-images.githubusercontent.com/35845239/231704791-66722939-3196-4793-8aa7-f6e86b11d0f9.mp4


endrilickollari (approved):
When there are notifications: 

image

When read all notifications: 

image

PR #2280, opened 2023-06-08
Fix segwit bestorders

smk762:
Closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2236
Fixes issue reported by @cipig in https://github.com/KomodoPlatform/atomicDEX-Desktop/pull/2230#discussion_r1155283698

To test:
- Login, go to dex pro page and select an order from the orderbook for a popular pair. 
- Wait for bestorders table to populate, then scroll through it
- You should see both the legacy and segwit version of a coin where supported (LTC or DGB should be there)
- Wait a minute or so without clicking anything to give the bestorders list time to refresh.
- Confirm both the legacy and segwit version are still present.
- Click on the segwit option of one of the coins in the bestorders list
- Confirm it changes the pair and loads a fresh orderbook as expected.
- Confirm doing a trade after this correctly sends or receives funds from the segwit address.

image
image

**Note**: Some crashes in the past were a result of an assert statement or different number of items in `m_model_data` and `m_orders_id_registry` - this includes where the legacy and segwit orders which had the same UUID (shared orderbook since https://github.com/KomodoPlatform/atomicDEX-API/issues/1598) and would omit one because uuid was an object key. Adding the `-segwit` suffix to the uuid prevents this. I have checked a lot of code to ensure this suffix has no unintended consequences.

This PR also sets the following coins to be activated by default to let users know about the segwit functionality:
- BTC-segwit
- LTC-segwit
- DGB-segwit

It might be an opportunity for some cross-promotion with he DGB team @tonymorony


endrilickollari (approved):
Confirm `LTC-segwit`, `DGB-segwit` and `BTC-segwit` are activated automatically: 

image

Everything looks okay here:

image

image

Also did a trade: 

image


OS: Windows 10

cipig (approved):

kivqa:
Checked on MacOS client. Segwit assets are displayed 
LTC-Segwit to USDT BEP20 swap order was completed successfully

PR #2290, opened 2023-06-10
Remove copy button for sensitive data

smk762:
Clipboard logging is a common way for malicious actors to target crypto users. With options like https://www.addictivetips.com/windows-tips/log-clipboard-entries-on-windows-10/ the attack doesn't need to be very technically capable.

The best defence against this attack is to disallow copying any sensitive data like private keys or seed phrases to the clipboard.

To Test:
- Launch, login
- Go to settings > Security >View seed & keys
- See only QR code button and no copy button alongside seed phrase and private keys

### DIFF
--- a/atomic_defi_design/Dex/Components/DefaultModal.qml
+++ b/atomic_defi_design/Dex/Components/DefaultModal.qml
@@ -12,8 +12,8 @@ Popup
     property int radius: 18
 
     anchors.centerIn: Overlay.overlay
-    horizontalPadding: 88
-    verticalPadding: 52
+    horizontalPadding: 40
+    verticalPadding: 25
 
     modal: true
     focus: true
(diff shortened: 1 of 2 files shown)


endrilickollari (approved):
Looks good in Windows 10, 
Can copy address but no copy button for priv key: 

image

cipig:
So how are we supposed to use the privkeys? Input every single char one by one?

tonymorony:
if user got malware so bad that it's control OS clipboard I think it can just make screenshot/record screen with such level of permissions, so I'm not sure if this change makes app much more secure, but definitely adds more hustle with manual privkeys typing and etc. 🤔

smk762:
I was thinking about adding setting for "paranoid mode / convenience mode" which could implement this and some other things. 
Added this PR based on chat in the `opsec`channel during mobile testing, but we can leave this PR open for discussion and omit from 0.5.8 until we figure out the best option

tonymorony:
I never seen implementation without copy button  🤔
e.g. metamask (I like their additional warning about looking into screen and etc.): 

image

Screenshot 2023-06-12 at 12 22 42

smk762:
We could add a similar warning (and also about clipboard). The best balance I think is keeping user as safe as possible without making it hard to use. Might b an idea to add an eyeball icon to each also so not everything is shown on screen all at once. That said, its trivial to convert wifs.

PR #2292, opened 2023-06-11
Update windows installer

smk762:
Closes https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/2108

- Incremented version
- created DEX_VERSION cmake variable
- Change windows installer path to include version

This will overcome "directory exists" errors when a user updates versions, but they will still see the error if they attempt to reinstall a version which is already installed.

Investigating this issue, it seems the QT Installer Framework does not intend for offline installers to do this - it is expected that the user run the maintainence tool to uninstall first.

There is a possible workaround - https://stackoverflow.com/questions/46455360/workaround-for-qt-installer-framework-not-overwriting-existing-installation

Ill give this a shot, but if it has problems, will submit this PR as a partial solution until a more comprehensive solution is ready.


endrilickollari (approved):
Looks good in Windows 10: 

Did the installation before, installed again warning message appeared: 

image

Did the installation anyway and looked good: 

image

kivqa:
@smk762 
Every installation in same folder displayed as uniqe instance in Control Panel > Programs > Programs and features  for Windows 10
If user uninstalls one of these instances then remained instances displayed incorrectly
telegram-cloud-photo-size-2-5206575539551324669-y

smk762:
Thanks for report - I'll see how that can be fixed. What happens when you try to uninstall one of the duplicates from add/remove programs?

kivqa:
The instances are remained in the program list without icons and it's possible to remove without any affect . But need to test behavior with different folders  for instances

PR #2335, opened 2023-08-02
update prices url

smk762:
To test:
- load app
- confirm prices data is available
- confirmed prices data is accurate


endrilickollari (approved):
Looks good in Windows 10: 


image

PR #2336, opened 2023-08-02
fix GLIBC error

cipig:
fix error with debian 11 and other older linux versions
```
AntaraAtomicDexAppDir/usr/bin/komodo-wallet: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: versionGLIBCXX_3.4.30' not found (required by AntaraAtomicDexAppDir/usr/bin/komodo-wallet)
```


smk762 (approved):

smk762:
This is a temporary solution. Longer term, we should move away from GHA environments an use our own docker container. ref https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2338

Issue #2314, opened 2023-06-23
[FR]: Overflow & background color enhancments


endrilickollari:
At "Add Crypto" modal, scrollable view, and overflow to be fixed because it reaches the point of input. I think the overflow and scrollable view would be perfect to be between "Select all assets" to "You can still enable ....", basically like overflow at "Swap Successful".

![image](https://github.com/KomodoPlatform/atomicDEX-Desktop/assets/32954585/bd7ded44-8405-4a22-a164-8291076f8aba)

![image](https://github.com/KomodoPlatform/atomicDEX-Desktop/assets/32954585/35abd5d4-0301-4f1e-87fe-feda1f19072a)

Swap Successful overflow: 

![image](https://github.com/KomodoPlatform/atomicDEX-Desktop/assets/32954585/38b1c6a5-227c-4c44-9975-68102ae28111)

Need to adjust the background color of tags with font color, there are times that the color of tag is as dark/bright as the font color and the user can't read the tag.

![image](https://github.com/KomodoPlatform/atomicDEX-Desktop/assets/32954585/5a2fb1d8-fb13-4b71-a26d-f83591b55978)

![image](https://github.com/KomodoPlatform/atomicDEX-Desktop/assets/32954585/3abd0dcb-a4a6-470a-b45f-bb7255dbc080)

PR #2349, opened 2023-09-09
Fix dex overflow

smk762:
Closes: https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2345

This PR also sets min height to 720px (down from 800px), which may introduce some overflows at the bottom of the app. Margins and components have been reduced to accommodate this, though some places may still need work. 

To test:
- Go to dex pro view
- Interact with all elements and confirm no overflows are seen.
- Check all pop up windows in app to confirm working and visually ok (including from outside dex page. E.g. adding coins, viewing orders etc)
- Check any columns where hidden components (e.g. error messages) may result in elements being pushed outside the bottom of the app window.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,13 +60,13 @@ endif ()
 ##! We fetch our dependencies
 if (APPLE)
     FetchContent_Declare(mm2
-            URL https://github.com/KomodoPlatform/komodo-defi-framework/releases/download/v1.0.5-beta/mm2-1d8bebd15-Darwin-Release.zip)
+            URL https://sdk.devbuilds.komodo.earth/main/mm2_79f6205-mac-x86-64.zip)
 elseif (UNIX AND NOT APPLE)
     FetchContent_Declare(mm2
-            URL https://github.com/KomodoPlatform/komodo-defi-framework/releases/download/v1.0.5-beta/mm2-1d8bebd15-Linux-Release.zip)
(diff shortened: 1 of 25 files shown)


cipig:
how about 1366x768 instead of 1280x720? it covers all other resolutions besides 1280x720

smk762:
> how about 1366x768 instead of 1280x720? it covers all other resolutions besides 1280x720. i wonder anyway who uses 1280x720... are there still monitors on Desktop with this low res? mine is 3840x2160 :-)

I'm on 3840x1920 :laughing: 

I chose 1280x720 as the absolute minimum to cover the widest range, there's about 5% of desktop users still with that res. Previously it was 1280*800 so this just reduces height 80px which was simple enough. Setting a different min width has a cascade of updates to do exp in the dex pro page.

The main purpose of this PR tho was to fix overflow on the dex pro page since  the sidebar became "always expanded". I made the sidebar a bit thinner, reduced margins in dex pro view components and fiddled with the chart until it was the right size. 

In future we might increase the min window

smk762:
Also closes: https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2314

kivqa (approved):
There is no overlapping or overflow for elements on dex  on my windows VM with screen resolution 1280x800
image
Actually resolution 1280x720 is TV standard know as HD ready. So it's really possible that some users can use their TVs instead of monitors. 
I did not reproduce issues described in #2314

PR #2352, opened 2023-09-09
Enhance fiat settings

smk762:
Closes: https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2307

To test:
- In settings, view "recommended fiats" row. Selected fiat should be brighter white, with other options greyed (and a slight lightening on mouse over).
- Select a different fiat from the drop down which is in the recommended row already. The highlighted option should change.
- Select a different fiat from the drop down which is not in the recommended row. It should be added to the start of the "recommended fiats" row, and be highlighted.
- Confirm selection is actually applied to app, and values are displayed in the selected fiat


cipig:
switched a bit back and forth and now i have USD 2 times in the recommended list
image
do we actually need to add everything to the recommended list?

kivqa:
same for me on windows. Recommended list always changes when user click on any currency: e.g. clicking few times between EUR and RUB 
image

smk762:
Thanks, this should be relatively easy to fix. Will push update later today

cipig (approved):

PR #2353, opened 2023-09-09
update addressbook tags

smk762:
Closes: https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2311

Addressbook tags are now limited to 10 chars and 5 per contact
image

A couple of buttons were renamed also for clarity.
image
image
image

To test:
- Add tags to a contact. You should not be able add more than 5 tags, or a tag longer than 10 letters. With all 5 tags at max length, there should be no overflows.

### DIFF
--- a/assets/themes/Binance - Dark/colors.json
+++ b/assets/themes/Binance - Dark/colors.json
@@ -122,7 +122,7 @@
   "listItemEvenBackground": "#161515",
   "listItemHoveredBackground": "#5B4704",
   
-  "addressBookTagColors": ["#627EEA", "#FFD87A", "#F7931A"],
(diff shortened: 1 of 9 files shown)


kivqa:
User can add only 5 tags max for one contact.
Max tag name length is 10 characters. 
There is no overflow for tags with max length. Tags displayed on two lines on Address Book screen:
image
But on Edit contact form user should scroll tags because 5 tags could not be displayed in one line.
image
Make sense to add 2 lines for tags on Edit contact form

kivqa:
As well there is small bug/feature. Cancel update button is not applicable for tags changes. If user added or removed some tags Cancel button does not discard these changes

kivqa (approved):
Approved. 
All targets of this PR are reached

PR #2355, opened 2023-09-11
Upgrade Z coin activation

smk762:
ref: https://github.com/KomodoPlatform/komodo-defi-framework/pull/1922
closes: https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2354

This PR introduces the `sync_params` feature in the newly released 1.0.7 API. This allows the user to specify which block to begin sync from - previously it was hardcoded as block 1900000 (for ARRR).

After this PR is merged it will use the block height selected by the user in settings, and where the setting does not exist (e.g. for zombie) it will use the checkpoint block from coins_config.json

image

To test:
- From a clean wallet, activate ARRR.
- Confirm progress % is sane. 
- Select an earlier block from the settings, then restart app and try to activate again. The sync block value should appear in logs like `"first_sync_block":{"actual":1500000,"is_pre_sapling":false,"requested":1500000}`. You dont need to wait for full sync (it takes a long time), just confirm in logs.
- From an older wallet, repeat the above steps.

Note: Balance from transactions prior to the sync block will not be visible / included.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,13 +60,13 @@ endif ()
 ##! We fetch our dependencies
 if (APPLE)
     FetchContent_Declare(mm2
-            URL https://github.com/KomodoPlatform/komodo-defi-framework/releases/download/v1.0.5-beta/mm2-1d8bebd15-Darwin-Release.zip)
+            URL https://sdk.devbuilds.komodo.earth/main/mm2_79f6205-mac-x86-64.zip)
 elseif (UNIX AND NOT APPLE)
     FetchContent_Declare(mm2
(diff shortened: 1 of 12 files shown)


borngraced:
will there be an option to pick a date too? like an interactive date picker. @smk762

smk762:
closing in favor of https://github.com/KomodoPlatform/komodo-wallet-desktop/pull/2356

smk762:
> will there be an option to pick a date too? like an interactive date picker. @smk762

heh, I thought the same, just wasnt sure how long it would take. Turns out, not so long.

PR #2360, opened 2023-09-15
Fix ATOM

smk762:
While working on the next RC, I noticed ATOM was failing to activate. Part of the problem was infra which was fixed earlier this week. The other part was a bad conditional in the legacy desktop codebase.

To test:
- Login to app
- Activate ATOM


cipig (approved):
image

PR #2370, opened 2023-09-22
fix tendermint tx history

smk762:
closes https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2368

To test:
- Log in to wallet with recent ATOM transactions
- activate ATOM, and open it in wallet page
- tx history should be visible
- Check tx history for ZHTLC coins also to confirm no regression.

Note: due to node pruning, older transactions may not be visible.


cipig (approved):
image
 and ARRR history also still present
image

kivqa (approved):
Approved.
ATOM txs are displayed correctly (old and new).
ARRR txs history does not have regression in this PR.

PR #2379, opened 2023-10-11
use default libtool

smk762:


kivqa (approved):
Approved. 
App builds for macOS works as expected. Checked on macos-13

PR #2383, opened 2023-11-05
Show trade volume for pair under pro orderbook

smk762:
Closes https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2382 and https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2384
image

To test:
- Confirm pairs match values at https://defi-stats.komodo.earth/api/v3/gecko/tickers
- Confirm reverse pair shows same values
- Confirm segwit / non-segwit pairs show same values
- If pair not in list on API, or zero volume traded, volume text should not display
- Confirm values update on fiat settings change as expected.
- Confirm pro view combo box text changes as expected when changing currency via menubar

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,13 +60,13 @@ endif ()
 ##! We fetch our dependencies
 if (APPLE)
     FetchContent_Declare(mm2
-            URL https://sdk.devbuilds.komodo.earth/dev/mm2_a9f606d-mac-x86-64.zip)
+            URL https://sdk.devbuilds.komodo.earth/main/mm2_79f6205-mac-x86-64.zip)
 elseif (UNIX AND NOT APPLE)
     FetchContent_Declare(mm2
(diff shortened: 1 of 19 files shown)


kivqa:
@smk762  trade volume works only for USD currency. If I use any other currency volume is zero
e.g. GBP 
image

smk762:
> @smk762 trade volume works only for USD currency. If I use any other currency volume is zero e.g. GBP image

Should be fixed now

https://github.com/KomodoPlatform/komodo-wallet-desktop/assets/35845239/3e5346fb-0c29-4fbb-ae7f-85ad3b79f823

kivqa (approved):
Approved.
Trade volume displayed correctly for different pairs and its matching with API response.
Changing currency in the app now handled correctly for trading volume.
There is not translation but its better to fix in separate issues with other places without translation.

cipig (approved):

PR #2386, opened 2023-12-10
Reduce high Tendermint fees

smk762:
Solves https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2371 until an API level solution is available.
To Test:
- try sending Tendermint assets (ATOM, IRIS, OSMO, ATOM-IRIS_IBC) without custom fee.
- try sending Tendermint assets (ATOM, IRIS, OSMO, ATOM-IRIS_IBC) with custom fee.

IRIS seemed to be the first to reject the tx when static fee was reduced further. The following values should work for custom fees for all Tendermint coins:
- gas_price: 0.05
- gas_limit: 150000

### DIFF
--- a/atomic_defi_design/Dex/Addressbook/AddAddressForm.qml
+++ b/atomic_defi_design/Dex/Addressbook/AddAddressForm.qml
@@ -31,6 +31,9 @@ Dex.Rectangle
             case "QRC-20":      return "QTUM"
             case "BEP-20":      return "BNB"
             case "ERC-20":      return "ETH"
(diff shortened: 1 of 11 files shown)


cipig (approved):
specifying custom fees for sending an IRIS tx works fine
when i use the values mentioned in this PR, the result is exactly the same as when i don't specify custom fees, which is expected because the values are the default ones
image

kivqa (approved):
Approved.
ATOM sending works as expected for default and custom fee. 
Fee reduced by 4-5 times.

PR #2397, opened 2023-12-21
Tweak mm2 proc

smk762:
- renames binary to avoid kill conflicts
- displays rpc port in settings > about


cipig (approved):

kivqa (approved):
Approved. 
mm2 proc:
image
port in settings:
image

PR #2414, opened 2024-03-28
rm paprika api services

smk762:
Replaces https://github.com/KomodoPlatform/komodo-wallet-desktop/pull/2413

To Test:
- Confirm tickers are returning correct price / balance data in all areas of the app
- Confirm fiat conversions are applied as expected


AndrewDelaney (commented):
There is an issue when switching currencies on the wallet page, where the balance of a coin is not updated correctly. The currency symbol changes, but the number does not.
This issue was not introduced, it existed before, but am not sure if this was meant to be addressed in this PR?

See video for details:

https://github.com/KomodoPlatform/komodo-wallet-desktop/assets/30876243/60e3ec7e-e7bc-4d49-a39c-deaeac379f7c

cipig (approved):

PR #2415, opened 2024-03-28
add pair trades 24hr stat

smk762:
Before:
image

After:
image

To test:
- Check a few random pairs in dex pro view.
- If pair has stats data at https://defi-stats.komodo.earth/api/v3/pairs/volumes_24hr it should display below the orderbook.
- If pair is not within the data from that endpoint, there have been no trades in the last 24 hours and there should be no stats shown under the orderbook.


AndrewDelaney (approved):
Tests were successful. The following items were tested:
- Comparing trades on DEX pro view with stats data from endpoint:
  - Number of trades matches
  - Trade Volume matches
  - No 24hr history underneath Order Book where no entry exists in stat data
- UI around Order Book still functional and styled neatly

PR #2420, opened 2024-04-01
Add/order info

smk762:
Closes https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2412
cc: @TheComputerGenie

To Test:
- Create an order which will not be matched (e.g. Sell 1 DOC for 10 MARTY).
- Wait for the order to revert to maker after 60 seconds.
- Click on the order to view its details
- See text for Max volume and Min volume in order details modal

image


AndrewDelaney (approved):
Testing successful. The following items were tested:
- Order that cannot be match is created and added to orders list
- Order reverts to maker after 60 seconds 
- Detailed view of order displays min and max volumes

PR #2421, opened 2024-04-01
open lcw when chart is clicked

smk762:
Closes: https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2418

To test:
- login and go to DEX tab
- select a pair which shows a chart
- click anywhere on the chart
- see https://www.livecoinwatch.com/ open in a tab in your default browser.

Known issues/limitations:
- As the chart code is generated by 3rd party JS, we cant just change how the hyperlink in the bottom left corner responds when clicked on, so instead the whole chart area is now "active" - you can click anywhere on the chart and it will open LCW.

### DIFF
--- a/atomic_defi_design/Dex/Exchange/ProView/Chart.qml
+++ b/atomic_defi_design/Dex/Exchange/ProView/Chart.qml
@@ -70,6 +70,7 @@ Item
                         transform: scale(${Math.min(scale_x, scale_y)});
                         transform-origin: top left;
                     }
+                    a { pointer-events: none; }
                 </style>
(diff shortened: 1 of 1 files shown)


AndrewDelaney (approved):
Testing successful.
Clicking on the graph for any valid coin will open Live Coin Watch website in browser.

It would be useful to have the graph open the website on the specific coin being viewed in Komodo Wallet, however, obtaining the correct URL for the individual coin might prove difficult dealing with a third party website; seems unnecessary for this update.

smk762:
Unfortunately the url format was not exactly the same as the api ids data we have to generate the chart.

PR #2454, opened 2024-06-13
Fix menubar currency change

smk762:
Closes: https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2444

This affects the price display on all pages of the app, wherever a non standard reference currency is selected via a click on the menubar.

To Test, select a range of fiat currencies (via settings) and reference currencies (via the menu bar) and navigate through the app pages to ensure all price info displayed is sane.


cipig (approved):
image
image
image

PR #2455, opened 2024-06-13
more triggers to scroll asks to bottom in pro orderbook

smk762:
DocAgnew on discord reported cases where the asks (in red) for the pro orderbook would not scroll to the bottom so the center point price in the bid/ask lists were tightened.

This PR adds a few additional triggers and timer delays to ensure this scroll is enforced once orderbook loads when 
- flipping pair
- changing pair
- changing buy/sell mode
- loading dex pro tab.

To test:
- login and go to pro screen. 
- select a pair with enough orders to extend the list beyond the window (e.g. KMD/LTC)
- flip the pair, change the pair, change buy to sell change tab and change back. For all of these the values in the top sectiopn oif the orderbook (in red) should scroll down fairly quickly after loading so that the values at the bottom of the red list are as close to the values at the top of the green list as possible.

For example: This is good
image

And this is not (unless manually scrolled after loading)
image

### DIFF
--- a/atomic_defi_design/Dex/Exchange/Trade/OrderBook/List.qml
+++ b/atomic_defi_design/Dex/Exchange/Trade/OrderBook/List.qml
@@ -72,16 +72,28 @@ Item
     }
 
     Connections {
-        target: API.app.trading_pg
+        target: API.app.trading_pg;
 
-        function onMarketModeChanged: {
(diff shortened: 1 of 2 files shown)


cipig (dismissed):

cipig:
fixes https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2145

PR #2462, opened 2024-07-30
Rename all instances of mm2 to kdf

smk762:
Closes https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2463

Basic smoke test needed:
- create/import wallet
- login
- send a transaction
- do a swap

Repeat the above with an existing wallet (created before this build)


AndrewDelaney (dismissed):
Testing was successful here. Couldn't find any issues during smoke tests.
The following processes were tested:
- Installation where previous data exists on environment
- Log in to existing wallet on environment
- Complete swap on existing wallet
- Clean installation
- Import wallet
- Create wallet
- Complete swap on imported wallet
- Complete swap on new wallet
- Recieve to imported wallet
- Send from imported wallet

cipig:
Problem is that directory with DB was renamed to kdf from mm2, so when you update you will loose the entire swap history (then failed swaps can't be refunded any more) and also the ARRR-DBs. Directory should be moved back to mm2 or we need a way to upgrade, like moving the directory including it's entire content.

ShantanuSharma9873:
Completed the Smoke Testing here, and everything looks fine. 

- Logged in to an existing Wallet 
- Completed a SWAP
- Imported a Wallet
- Logged in successfully 
- Completed a SWAP
- Created a New Wallet 
- Logged in successfully
- Completed a SWAP

smk762:
> Problem is that directory with DB was renamed to kdf from mm2, so when you update you will loose the entire swap history (then failed swaps can't be refunded any more) and also the ARRR-DBs. Directory should be moved back to mm2 or we need a way to upgrade, like moving the directory including it's entire content.

[Latest commit](https://github.com/KomodoPlatform/komodo-wallet-desktop/pull/2462/commits/1b146a4ddeff5a15269fadc91c787e23ee2dd706) should resolve this.

cipig (approved):
works fine now, mm2 dir is renamed to kdf and swap history retained

PR #2464, opened 2024-08-12
Update app version

smk762:
- Confirm version string in settings is 0.8.0
- Confirm VOTE2024 and all KIP voting chains have been removed


ShantanuSharma9873:
- The Application Version has been successfully updated to 0.8.0 beta. 
image

- But able to see the 'VOTE2024' at the Add Crypto. Also, the user is able to Enable the coin.
image

AndrewDelaney (commented):
Both the Vote 2024 and KIP tokens are still present.
What has changed, though, is that the Vote 2024 asset is no longer added by default.

smk762:
> Both the Vote 2024 and KIP tokens are still present. What has changed, though, is that the Vote 2024 asset is no longer added by default.

They will appear until removed from https://github.com/KomodoPlatform/coins/blob/master/coins#L15642
I'll do the coins repo PR now for this

smk762:
https://github.com/KomodoPlatform/coins/pull/1131

cipig (dismissed):

PR #2465, opened 2024-08-12
Fix/dexpro/orderbook/bignum alignment

smk762:
Closes https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2427

To test:
- login and activate MAZA
- go to pro view, and select KMD/MAZA for orderbook
- confirm no misalignment


cipig:
it now looks like this
image
pay attention to the bids... there is one for 30k MAZA and one for 300k MAZA... they look the same (because they have a different number of decimals)
the ask for 9k and the one for 99k also look the same
i fear we can't solve the problem by using different number of decimals for different entries in same view... we need to cut down the decimals for all entries to the same number or we need to make the column wider

AndrewDelaney:
Everything does look very nice and neatly aligned. The only issue I have is with the column headings, and the amount of space left unutilised. I agree with @cipig, the columns can be made wider. I have created a proposed idea for an improved layout that may remedy the issue @cipig raised.

| Before | After |
| --- | --- | 
| OrderBookAlignment | OrderBookAlignment(Proposal) |

This layout will help show more of the number and make it easier to compare. I know we round to 8 decimal places, but perhaps we should round down to the last non-zero digit within the 8 decimal places? Then right aligning all the columns to make comparisons easier.
e.g.
70.540000000085 -> 70.54
45.00000000 -> 45.0
97.6800430000091 -> 97.680043

Just an idea, though this is a lot more work than simple alignment, and we'd have to look at changing it across the board.

cipig:
> This layout will help show more of the number and make it easier to compare. I know we round to 8 decimal places, but perhaps we should round down to the last non-zero digit within the 8 decimal places?

the values in the columns should be right-aligned, not left-aligned, so you can see right away which value is bigger... and you can only do that if all values have the same number of decimals, if they differ it does not work

ShantanuSharma9873:
- I’m unable to load the wallet. I downloaded the Windows build - [komodo-wallet-windows-release-installer-fb1f352.exe](https://github.com/KomodoPlatform/komodo-wallet-desktop/actions/runs/10348976108/artifacts/1801018959) and tried logging into an existing wallet and creating a new one, but the loader keeps rotating, and I’m unable to access the wallet. 
https://github.com/user-attachments/assets/84a24f00-1a06-4e22-84c8-868681dccb8f

- Aside from this issue, I agree with the points raised by @cipig and @AndrewDelaney . Adding a divider would help differentiate sections, and right-aligning the Quantity and Total columns would make the numbers more readable.

cipig:
if we need to save space, we could reduce the number of decimals for "Quantity" to 6, in general, for all coins/orderbooks... for BTC (the coin with highest value in USD) the min. volume in a swap is anyway 0.00777000, so orders with lower volumes can't be posted

atm this change makes it worser, since it "damages" orderbooks that previously looked fine, like DOGE/KMD:
image

cipig:
just realized that we already had it to 6 decimals for "Quantity" and "Total"
this is how DOGE/KMD orderbook looks like without this PR:
image
and MAZA/KMD:
image
the big orders seen in https://github.com/KomodoPlatform/komodo-wallet-desktop/issues/2427#issue-2233330136 are gone in the meantime, so the problem is only with big numbers, then it gets unaligned... i fear the only way to solve it is to make tab wider

cipig (approved):
much better, thanks
here is one example with big values... BTTC-BEP20/KMD
image
only problem i see is that it uses different numbers of decimals... the most expensive ask has 4 decimals in "Quantity", the others have 3
but such cases are now pretty rare

PR #2482, opened 2024-09-21
Disable/time sync check

smk762:
This feature has been causing problems for some users, so it will be disabled until a better solution is found

### DIFF
--- a/src/core/atomicdex/services/sync/timesync.checker.service.cpp
+++ b/src/core/atomicdex/services/sync/timesync.checker.service.cpp
@@ -42,6 +42,7 @@ namespace
 
     bool get_timesync_info_rpc(web::http::http_response resp_http)
     {
+        return true;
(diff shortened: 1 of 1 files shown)


ShantanuSharma9873 (commented):
The '**Time Sync**' popup no longer appears initially, but after changing the system's time, it is observed that SWAPs start failing, and a popup with the message **'Selected Order Removed'** (as shown in the image) is displayed. When the system's time is corrected, the SWAPs proceed successfully again.

**NOTE:** The same popup appears again during an ongoing swap.
image

**Swaps:**
image

Tested on Windows 11.

cipig:
> but after changing the system's time, it is observed that SWAPs start failing

if your system time is not in sync (> 60 seconds) all your swaps fail

cipig (approved):

Issue #2505, opened 2025-01-08
Remove trading bot toggle


smk762:
**To Reproduce**
1. Login 
2. Go to settings
3. See `enable trading bot` toggle
4. press the toggle and wonder why it is there if the dex tab is removed.

**Expected behavior**
While DEX tab is removed, toggle should also be removed.

**Found in**
Current prod

smk762:
closing as complete

PR #2506, opened 2025-01-08
Remove trading bot setting

naezith:
Solves https://github.com/KomodoPlatform/komodo-wallet/issues/2505

image


naezith:
The `kIsWalletOnly` variable is not implemented in this repository yet; that's why it fails.

PR #2510, opened 2025-01-13
HD Seed Import

naezith:
This also adds `WalletType.hdwallet`; please review the implementation. It's not listed in the connect list like other `WalletType` entries so there's some custom code around that.

**Note:** Custom seed importing was broken in the KDF/SDK, so those parts were not tested.

Please test deeply.


lib/model/wallet.dart
```diff
@@ -133,6 +134,7 @@ class WalletConfig {
 enum WalletType {
   iguana,
   // TODO! add HD wallet type
```
CharlVS, line 136:
```suggestion
```

lib/views/wallets_manager/widgets/wallet_import_by_file.dart
```diff
@@ -51,6 +57,10 @@ class _WalletImportByFileState extends State<WalletImportByFile> {
     return _filePasswordError == null;
   }
 
+  bool get _isButtonEnabled {
+    return _eulaAndTosChecked;
+  }
```
CharlVS, line 62:
```suggestion
  bool get _isButtonEnabled => _eulaAndTosChecked;
```

smk762:
- confirmed non-BIP39 seed import from file rejected if HD mode selected, or `I understand` not input.
- confirm no regression when attempting wallet import with same as existing name
- confirm import of 24 word seed via input in HD mode functions as expected.
- Error seen in tx history panel `something went wrong` when zero balance. Once funded, functioning as expected.
- Error seen when clicking on wallet>receive button
image
- HD addresses list shows balances, but combined balance still zero
Screenshot from 2025-01-15 14-24-25
- KMD rewards claim button leads to endless spinner and error in console logs
image
- Address creation and "hide addresses with zero balance" functioning as expected.

Other errors seen in console logs:
image

ShantanuSharma9873 (commented):
Tested the PR, and here are my observations from the smoke testing:

- Successfully created a new wallet in both modes: With HD Wallet and Without HD Wallet.
- Successfully saved the seed phrase and downloaded the seed file.

HD Wallet 1
HD Wallet 2

**Non-HD Wallet Mode:**
- The "Create Address" button was disabled.
- Hovering over the button displayed the tooltip: _"Current wallet mode does not support multiple addresses."_

**HD Wallet Mode:**
- Successfully created multiple addresses for a single coin.
- Verified that the addresses could be copied and their QR codes were accessible.

**Additional Observations:**
**Wallet Name Missing:**
- In 2/5 cases, the wallet name was not visible at the top right corner, causing the logout button to disappear.
- Had to log out via the Settings tab as a workaround.
Video: 
https://github.com/user-attachments/assets/f6d92e94-d09e-4b05-88d6-f2388f425e08

**Seed File Import Issue:**
- When uploading the seed file via the **"Import Wallet"** option, the message _"Wallet name exists"_ was displayed.
- The seed file being uploaded was the same one downloaded via the "Download Seed File" option in the Settings tab.
Video:
https://github.com/user-attachments/assets/980d5595-57ad-4e83-8d4c-45049b384710

The testing is still left to be completed.

CharlVS:
Thank you for your thorough review, @ShantanuSharma9873. Please continue any further findings here: #2516

PR #2521, opened 2025-02-05
fix(router): frozen layout on navigation and resize

takenagain:
Changes

- bump SDK version hash in `pubspec.lock` (`dev` branch is using an outdated SDK release)
- fixes the frozen layout when navigating and resizing
- fixes blank coin charts on the guest view (CEX asset not found in available coins list)

### DIFF
--- a/lib/bloc/cex_market_data/price_chart/price_chart_bloc.dart
+++ b/lib/bloc/cex_market_data/price_chart/price_chart_bloc.dart
@@ -33,8 +33,10 @@ class PriceChartBloc extends Bloc<PriceChartEvent, PriceChartState> {
       if (state.availableCoins.isEmpty) {
         final coins = (await cexPriceRepository.getCoinList())
             .where((coin) => coin.currencies.contains('USDT'))
-            // filter out assets that are not in the known coins list to 
-            // prevent errors when using the SDK with the `.single` property
+            // `cexPriceRepository.getCoinList()` returns coins from a CEX
+            // (e.g. Binance), some of which are not in our known/available
+            // assets/coins list. This filter ensures that we only attempt to
+            // fetch and display data for supported coins
             .where((coin) => sdk.assets.assetsFromTicker(coin.id).length == 1)
             .map((coin) async {
           double? dayChangePercent = coinPrices[coin.symbol]?.change24h;
--- a/lib/bloc/cex_market_data/price_chart/price_chart_bloc.dart
+++ b/lib/bloc/cex_market_data/price_chart/price_chart_bloc.dart
@@ -37,7 +37,7 @@ class PriceChartBloc extends Bloc<PriceChartEvent, PriceChartState> {
             // (e.g. Binance), some of which are not in our known/available
             // assets/coins list. This filter ensures that we only attempt to
             // fetch and display data for supported coins
-            .where((coin) => sdk.assets.assetsFromTicker(coin.id).length == 1)
+            .where((coin) => sdk.assets.assetsFromTicker(coin.id).length >= 1)
             .map((coin) async {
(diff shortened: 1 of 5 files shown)


lib/bloc/cex_market_data/price_chart/price_chart_bloc.dart
```diff
@@ -30,6 +33,9 @@ class PriceChartBloc extends Bloc<PriceChartEvent, PriceChartState> {
       if (state.availableCoins.isEmpty) {
         final coins = (await cexPriceRepository.getCoinList())
             .where((coin) => coin.currencies.contains('USDT'))
+            // filter out assets that are not in the known coins list to 
+            // prevent errors when using the SDK with the `.single` property
+            .where((coin) => sdk.assets.assetsFromTicker(coin.id).length == 1)
```
CharlVS, line 38: My intention with using the `.single` property instead of `.first`/`.firstOrNull` was to intentionally throw an exception if there is more than one asset for a given ticker. This was possible before the SLP project was abandoned.

Although unlikely, if the coins repo has a change pushed to it where this assumption no longer holds true, then we want to error out because there are many instances where we've assumed a given ticker will have exactly 1 or 0 coins/assets, which may wreak havoc if it's not treated as a breaking change.

Unless the surrounding context makes it redundant, please change this to:
```suggestion
            // filter out assets that are not in the known coins list to 
            // prevent errors when using the SDK with the `.single` property
            .where((coin) => sdk.assets.assetsFromTicker(coin.id).length >= 1)
```


Also, we could add an SDK method for `sdk.assets.maybeAssetsFromTicker`

lib/router/navigators/app_router_delegate.dart
```diff
@@ -37,7 +37,9 @@ class AppRouterDelegate extends RouterDelegate<AppRoutePath>
           materialPageContext = context;
           return GestureDetector(
             onTap: () => runDropdownDismiss(context),
-            child: const MainLayout(),
+            child: MainLayout(
+              key: ValueKey('${isMobile}_${routingState.selectedMenu}'),
```
CharlVS, line 41: Does the issue remain if `isMobile` is removed? Unless necessary, prefer excluding `isMobile` from the key since we want the responsiveness and adaptability to be as seamless as possible.

takenagain: Removing the `isMobile` flag does not appear to affect navigation or routing, so I removed it in [8ff69c2](https://github.com/KomodoPlatform/komodo-wallet/pull/2521/commits/8ff69c2f3dfdcde71c41f9e5aa0886a1255c286c)

lib/bloc/cex_market_data/price_chart/price_chart_bloc.dart
takenagain: The `await cexPriceRepository.getCoinList()` function on line 34 above returns a list of coins from Binance, some of which are not in our coins repository / config. Line 38 filters out assets that are not in the “known”/available assets list, or that might have more than one asset for a given ticker.

The entire application crashes in the guest/unauthenticated Wallet page because of the `.single` call in [getSdkAsset](https://github.com/KomodoPlatform/komodo-wallet/blob/73ac3630f5cf849e40ca7af154720060afebcc5d/lib/shared/utils/utils.dart#L682) throwing an exception in the [UI layer](https://github.com/KomodoPlatform/komodo-wallet/blob/fix/frozen-layout/lib/views/wallet/wallet_page/charts/coin_prices_chart.dart#L54) (no entries in the set).

I updated the comment to clarify the purpose of the filter in [8ff69c2](https://github.com/KomodoPlatform/komodo-wallet/pull/2521/commits/8ff69c2f3dfdcde71c41f9e5aa0886a1255c286c)

CharlVS (approved):

PR #2527, opened 2025-02-12
fix(coins-bloc): coin deactivation not triggering state change and Bloc rebuild

takenagain:
Fix coin deactivation not triggering UI rebuilds reported by @CharlVS, along with the review suggestions from Claude.


CharlVS (approved):
Thank you Kanye, very cool!

PR #2565, opened 2025-03-07
chore: add vscode config directory

takenagain:
Simplifies the following use-cases:
- Initial project setup
- [Git worktrees workflow](https://www.gitkraken.com/learn/git/git-worktree) - simplifies juggling multiple branches

Includes the `--web-port=` flag to simplify testing (login rather than register/restore)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,7 @@ sdk/
 # The .vscode folder contains launch configuration and tasks you configure in
 # VS Code which you may wish to be included in version control, so this line
 # is commented out by default.
-.vscode/
+# .vscode/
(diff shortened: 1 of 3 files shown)


smk762:
Thanks - looks useful. Can you please add some info to the repo docs detailing the how/why of the `--web-port=` flag, and any other useful/missing info for VS Code env setup.

CharlVS:
There are some changes I'd like to make to this, which I will get to after I'm up-and-running. I'm marking this as a draft for the meantime so that it isn't merged while I'm away and to indicate that it's not ready to be reviewed.

PR #2569, opened 2025-03-20
feat(feedback): add api key management via envied and .env files

takenagain:
Adds the [ENVied](https://pub.dev/packages/envied) package to manage the Trello API keys via .env files. The main benefit that ENVied has over `--dart-define` is obfuscation of the API key in the final binary, making it slightly more difficult to decode, although it is still possible.

Possible issues with this implementation:
- Client-side API keys are accessible and should be considered public, even with obfuscation enabled. Permissions have to be configured with this in mind.
- Developers will be required to run `dart run build_runner build` when opening the project for the first time and whenever the `.env` file is modified.
  - The environment variable values are stored inside the generated dart file, so build_runner is necessary to update the values in the generated file

See the ClickUp comparison doc for further information: https://app.clickup.com/9015263475/docs/8cnm07k-5195.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,4 @@
+TRELLO_API_KEY=your_api_key
+TRELLO_TOKEN=your_token
+TRELLO_BOARD_ID=your_board_id
+TRELLO_LIST_ID=your_list_id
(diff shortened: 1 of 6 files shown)


takenagain:
@coderabbitai review

takenagain:
Closed in favour of revisiting at a later stage and possibly using another approach (i.e. firebase functions)

PR #2580, opened 2025-03-29
feat(auth): SDK password update migration

CharlVS:
Re-implement password update functionality using the recently added KDF RPC. This is necessary because the Flutter app no longer handles authentication and user accounts and is managed by KDF (exposed through the SDK)


smk762:
To Test:
- create or import a wallet
- change the password
- log out
- confirm login with old password fails
- confirm login with new password succeeds
- export wallet to file
- log out
- clear cache, and import recently exported wallet file
- confirm old password fails
- confirm new password works


@CharlVS please add any extras I've not covered.

smk762:
Ambiguous error message:
image

This was triggered when I input an incorrect "old password".

Other validation cases for matching / compliant "new password / confirm new password" inputs function as expected with informative error messaging.

The heading for this step implies an action not yet completed. Suggest change to `Importing seed file`
image

I've opened a [separate issue](https://github.com/KomodoPlatform/komodo-wallet/issues/2600) for allowing import of files with existing wallet name to allow `rename`.

CharlVS:
> The heading for this step implies an action not yet completed. Suggest change to `Importing seed file`

@TazzyMeister Please address this in your PR #2587

CharlVS:
> To Test:
> 
> * create or import a wallet
> * change the password
> * log out
> * confirm login with old password fails
> * confirm login with new password succeeds
> * export wallet to file
> * log out
> * clear cache, and import recently exported wallet file
> * confirm old password fails
> * confirm new password works
> 
> @CharlVS please add any extras I've not covered.

@smk762 that looks good. It goes beyond everything needed, but it's better to be safe since a funky state in KDF after password updates could be disastrous. The GUI changes are limited to the re-introduction of password updates (which is not stored by KW), so anything beyond that is testing for KDF.

smk762 (approved):
LGTM, thanks!

AndrewDelaney:
When attempting password change, if I set all three values to be the same new password I'm trying to set, this is the error message I reveive:
image

In this case it should be an "Incorrect current password" error message

PR #2584, opened 2025-03-31
fix(coins-bloc): coins re-activating after being disabled by the user

takenagain:
Fixes coins re-appearing in the coins list after being disabled (after a delay), as seen in #2526 

## Summary by CodeRabbit

- **Refactor**
  - Streamlined coin management to fetch current wallet data in real time, ensuring up-to-date coin status.
  - Enhanced responsiveness in coin activation, deactivation, and refresh operations through asynchronous updates.

### DIFF
--- a/lib/bloc/coins_bloc/coins_bloc.dart
+++ b/lib/bloc/coins_bloc/coins_bloc.dart
@@ -60,9 +60,6 @@ class CoinsBloc extends Bloc<CoinsEvent, CoinsState> {
   Timer? _updatePricesTimer;
   Timer? _reActivateSuspendedTimer;
 
-  // prevents RPC spamming on startup & previous inconsistencies with sdk wallet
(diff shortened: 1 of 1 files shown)


takenagain:
@coderabbitai review

CharlVS (approved):

ShantanuSharma9873:
The **Enable/Disable** feature for coins is working as expected. I did not encounter any issues during testing. After disabling specific coins and logging back into the wallet, I observed that the disabled coins no longer appear on the Wallet page or under the graph, which is the expected behavior.

However, please note that this also applies to the **KMD** coin. Once **KMD** is disabled and the user logs in again, it does not reappear on the Wallet page unless re-enabled.

Additionally, although it falls outside the current testing scope, I noticed that the test coins **(Doc & Marty)** were not visible even when the '**Enable Test Coin**' toggle was turned on in the settings.

PR #2598, opened 2025-04-11
fix(custom-token-import): new required parameter in asset constructor

takenagain:
Fixes the build error introduced in #2597


smk762 (approved):
Thanks for the quick fix!

PR #2576, opened 2025-03-25
Made Validation for Confirm Password Eager

TazzyMeister:
The password to confirm could be bypassed, now it will show an error that if it has not been filled in and it will show while the user is putting in their input

## Summary by CodeRabbit

- **New Features**
  - Enhanced password confirmation validation with clearer error messages for empty and mismatched confirmations.
  - Improved form input fields with new options for customizing validation behavior during user interaction.

- **Bug Fixes**
  - Updated translation messages to remove outdated feedback-related entries and add a specific error message for empty password confirmation.

PR #2607, opened 2025-04-29
fix(password): bypass password validation issue

TazzyMeister:
This PR replaces the previous PR [#2576] due to conflict issues.

The password to confirm could be bypassed, now it will show an error that if it has not been filled in and it will show while the user is putting in their input.


smk762 (approved):
Confirmed functional on wallet create and settings > password update.

PR #2740, opened 2025-06-03
Fix failing validation on trezor login

smk762:
A simple fix for a a significant problem.
- Updates trezor hidden password to required complexity to pass validation.
- Ensures automatically activated coins are registered in the GUI on Trezor Init.

### DIFF
--- a/lib/shared/utils/password.dart
+++ b/lib/shared/utils/password.dart
@@ -1,84 +1,9 @@
-import 'dart:math';
+import 'package:komodo_defi_types/komodo_defi_type_utils.dart';
 
-String generatePassword() {
-  final List<String> passwords = [];
-
-  final rng = Random.secure();
-
-  const String lowerCase = 'abcdefghijklmnopqrstuvwxyz';
-  const String upperCase = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ';
-  const String digit = '0123456789';
-  const String punctuation = '*.!@#\$%^(){}:;\',.?/~`_+\\-=|';
-
-  final string = [lowerCase, upperCase, digit, punctuation];
-
-  final length = rng.nextInt(24) + 8;
-
-  final List<String> tab = [];
-
-  while (true) {
-    // This loop make sure the new RPC password will contains all the requirement
-    // characters type in password, it generate automatically the position.
-    tab.clear();
-    for (var x = 0; x < length; x++) {
-      tab.add(string[rng.nextInt(4)]);
-    }
-
-    if (tab.contains(lowerCase) &&
-        tab.contains(upperCase) &&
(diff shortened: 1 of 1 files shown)


lib/bloc/trezor_init_bloc/trezor_init_bloc.dart
```diff
@@ -273,7 +274,7 @@ class TrezorInitBloc extends Bloc<TrezorInitEvent, TrezorInitState> {
   /// into a static 'hidden' wallet to init trezor
   Future<void> _loginToTrezorWallet({
     String walletName = 'My Trezor',
-    String password = 'hidden-login',
+    String password = 'Hidden-login1!',
```
CharlVS: Does the password have to remain constant between restarts? If not, rather opt for using `SecurityUtils.generatePasswordSecure(16)`, which generates a secure password that meets KDF's password criteria using the device's secure RNG.

CharlVS: @smk762

lib/bloc/trezor_init_bloc/trezor_init_bloc.dart
```diff
@@ -274,8 +275,9 @@ class TrezorInitBloc extends Bloc<TrezorInitEvent, TrezorInitState> {
   /// into a static 'hidden' wallet to init trezor
   Future<void> _loginToTrezorWallet({
     String walletName = 'My Trezor',
-    String password = 'Hidden-login1!',
+    String? password
   }) async {
+    password ??= generatePassword();
```
CharlVS, line 280: @smk762 I see this function isn't referenced anywhere else in the codebase. Could you please remove the reference and delete the function? Then instead reference `SecurityUtils.generatePasswordSecure(16);`

smk762:
```
lib/bloc/trezor_init_bloc/trezor_init_bloc.dart:276:23: Error: Undefined name 'SecurityUtils'.
    String password = SecurityUtils.generatePasswordSecure(16),
                      ^^^^^^^^^^^^^
lib/bloc/trezor_init_bloc/trezor_init_bloc.dart:276:37: Error: Method invocation is not a constant expression.
    String password = SecurityUtils.generatePasswordSecure(16),
                                    ^^^^^^^^^^^^^^^^^^^^^^
```
Searching codebase for `generatePasswordSecure` or `SecurityUtils` returned no results, but I did find this, which allowed app to build - https://github.com/KomodoPlatform/komodo-wallet/blob/main/lib/shared/utils/password.dart#L3

If you can suggest the required import for the function you mention, happy to use it.

CharlVS (approved):

PR #2764, opened 2025-06-11
test: switch SDK to `test/android-hostname-resolution`

takenagain:


smk762:
Confirmed working as expected here. Thanks for the quick fix!

PR #2796, opened 2025-06-18
Show Git Commit Hash on Settings › General Page

DeckerSU:
image

This PR adds the current Git commit hash to the “Settings › General” screen in the app. The commit hash is displayed alongside the existing app version and API version, allowing developers and QA to quickly identify exactly which source revision a given build was produced from.

### DIFF
--- a/.docker/build.sh
+++ b/.docker/build.sh
@@ -40,13 +40,13 @@ docker build $PLATFORM_FLAG -f .docker/komodo-wallet-android.dockerfile . -t kom
 # Create the build directory ourselves to prevent it from being created by the Docker daemon (as root)
 mkdir -p ./build
 
-GIT_COMMIT=$(git rev-parse --short HEAD | cut -c1-7)
+COMMIT_HASH=$(git rev-parse --short HEAD | cut -c1-7)
 
 ENV_ARGS=""
 ENV_VARS="GITHUB_API_PUBLIC_READONLY_TOKEN TRELLO_API_KEY \
 TRELLO_TOKEN TRELLO_BOARD_ID TRELLO_LIST_ID \
 FEEDBACK_API_KEY FEEDBACK_PRODUCTION_URL FEEDBACK_TEST_URL \
-GIT_COMMIT"
+COMMIT_HASH"
 
 for VAR in $ENV_VARS; do
   case "$VAR" in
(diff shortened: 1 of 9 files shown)


DeckerSU:
I replaced all references to `GIT_COMMIT` with `COMMIT_HASH`, as `COMMIT_HASH` sounds clearer in my opinion.

The only concern I have is that all the build workflows use `./.github/actions/generate-assets` to fetch packages and generate assets. That means `flutter build ...` is invoked once during asset generation and then again during the actual build step.

I’m thinking it might not be necessary to modify the `build_command` in the `matrix` with `--dart-define=COMMIT_HASH="$COMMIT_HASH"` since `COMMIT_HASH` is already defined in the generate-assets action [[here](https://github.com/KomodoPlatform/komodo-wallet/blob/b4d012e1ff1c8a13845197e428ca35e1e6d71a35/.github/actions/generate-assets/action.yml#L76)](https://github.com/KomodoPlatform/komodo-wallet/blob/b4d012e1ff1c8a13845197e428ca35e1e6d71a35/.github/actions/generate-assets/action.yml#L76).

With the current setup, we might end up with two `--dart-define=COMMIT_HASH` values, like:

```
flutter build --dart-define=COMMIT_HASH="$COMMIT_HASH" --dart-define=COMMIT_HASH="$COMMIT_HASH"
```

First from the matrix command and second from `generate-assets`. Then again in the actual build:

```
flutter build --dart-define=COMMIT_HASH="$COMMIT_HASH"
```

I think I should fix that to avoid redundant or conflicting defines.

DeckerSU:
Looks like it works now. But until now, our `generate_assets` action has always run:
```
COMMIT_HASH=$(git rev-parse --short HEAD)
```
right after checkout. On a `pull_request` trigger, GitHub performs a “merge-commit” of the PR branch into the base branch, so HEAD actually pointed at that synthetic merge commit—not at the real tip of the feature branch.

Will change that behavior ...

DeckerSU:
It looks like everything is working fine now. In the [preview](https://github.com/KomodoPlatform/komodo-wallet/pull/2796#issuecomment-2984747601), it correctly shows the actual commit of the PR HEAD, as expected, instead of the hash of the synthetic merge commit.

smk762 (approved):
Thanks! This is a very useful feature for QA which came in handy on legacy desktop. Related issue added to expand on this one https://github.com/KomodoPlatform/komodo-wallet/issues/2803

PR #2816, opened 2025-06-24
linux: call XInitThreads() to make X11 access thread-safe

DeckerSU:
Flutter’s engine spawns multiple threads (UI, raster, etc.). Under X11 these threads may issue X calls concurrently, which is undefined unless XInitThreads() is invoked before the first X request.

Without this, the app aborts on launch with:

```
  [xcb] Unknown request in queue while dequeuing
  [xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
  KomodoWallet: xcb_io.c:175: dequeue_pending_request: Assertion `!xcb_xlib_unknown_req_in_deq' failed.
```

This patch:

* adds `#include <X11/Xlib.h>`
* calls `XInitThreads()` at the very start of `main()` in `linux/runner/main.cc`
* ensures libX11 is linked via CMake

Wayland sessions were already stable; this change fixes crashes when running under pure X11 or XWayland.

- https://github.com/flutter/flutter/issues/170937
- https://github.com/flutter/flutter/issues/169470

### DIFF
--- a/linux/CMakeLists.txt
+++ b/linux/CMakeLists.txt
@@ -52,6 +52,7 @@ add_subdirectory(${FLUTTER_MANAGED_DIR})
 
 # System-level dependencies.
 find_package(PkgConfig REQUIRED)
+find_package(X11 REQUIRED)
 pkg_check_modules(GTK REQUIRED IMPORTED_TARGET gtk+-3.0)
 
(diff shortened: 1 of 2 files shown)


smk762 (approved):
Built and running on Ubuntu24 with `XDG_SESSION_TYPE: x11` for over an hour without any stability issues.

PR #2818, opened 2025-06-24
disable Firebase Analytics on Linux and explicitly handle Unsupported…

DeckerSU:
…Error in init

* Add a `Platform.isLinux` guard to immediately disable analytics on unsupported platforms
* Wrap `Firebase.initializeApp` in its own `try/catch` to catch `UnsupportedError` and short-circuit initialization
* Complete the `_initCompleter` with an error on unsupported platforms to unblock awaiting callers
* Preserve existing retry logic for other initialization failures without retrying when unsupported
* Log debug messages when initialization is skipped or fails due to lack of Linux support

### DIFF
--- a/lib/bloc/analytics/analytics_repo.dart
+++ b/lib/bloc/analytics/analytics_repo.dart
@@ -1,6 +1,7 @@
 import 'dart:convert';
 import 'dart:async';
-import 'dart:io';
+import 'package:flutter/foundation.dart'
+    show defaultTargetPlatform, kIsWeb, TargetPlatform;
 
 import 'package:firebase_analytics/firebase_analytics.dart';
 import 'package:firebase_core/firebase_core.dart';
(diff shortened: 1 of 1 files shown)


lib/bloc/analytics/analytics_repo.dart
```diff
@@ -1,5 +1,6 @@
 import 'dart:convert';
 import 'dart:async';
+import 'dart:io';
```
CharlVS: Please instead use `TargetPlatform` and `kIsWeb` from `package:flutter/foundation.dart`, similar to how it's done in `lib/firebase_options.dart` because `dart:io` isn't supported on web. I'm not sure why it builds and runs.

CharlVS (changes requested):
Thank you for your contributions.

lib/bloc/analytics/analytics_repo.dart
DeckerSU: Done in [https://github.com/KomodoPlatform/komodo-wallet/pull/2818/commits/9b1434e598903eb96df7e253c7da0f84352c0d15](https://github.com/KomodoPlatform/komodo-wallet/pull/2818/commits/9b1434e598903eb96df7e253c7da0f84352c0d15). Thanks for sharing best practices.

PR #2820, opened 2025-06-25
Fix null my_info handling for swaps

CharlVS:
## Summary
- handle `null` `my_info` when parsing swaps
- make `Swap.myInfo` nullable throughout the app
- guard UI and sort logic when `myInfo` is absent

## Testing
- `flutter pub get --offline --enforce-lockfile` *(fails: version solving failed)*
- `flutter analyze --no-pub` *(fails: analysis server exit)*
- `dart format -o none lib/blocs/trading_entities_bloc.dart lib/model/swap.dart lib/views/dex/dex_helpers.dart lib/views/dex/entities_list/history/history_item.dart lib/views/dex/entities_list/history/swap_history_sort_mixin.dart lib/views/dex/entities_list/in_progress/in_progress_item.dart lib/views/dex/entities_list/in_progress/in_progress_list.dart lib/views/dex/entity_details/swap/swap_details.dart lib/views/dex/entity_details/swap/swap_details_step_list.dart`

------
https://chatgpt.com/codex/tasks/task_e_685bf1b20d988326ae4dce966bebd0e0

### DIFF
--- a/lib/blocs/trading_entities_bloc.dart
+++ b/lib/blocs/trading_entities_bloc.dart
@@ -55,7 +55,8 @@ class TradingEntitiesBloc implements BlocBase {
   List<Swap> get swaps => _swaps;
   set swaps(List<Swap> swapList) {
     swapList.sort(
-        (first, second) => second.myInfo.startedAt - first.myInfo.startedAt);
+        (first, second) =>
+            (second.myInfo?.startedAt ?? 0) - (first.myInfo?.startedAt ?? 0));
     _swaps = swapList;
     _inSwaps.add(_swaps);
   }
@@ -126,7 +127,7 @@ class TradingEntitiesBloc implements BlocBase {
     final double swapFill = swaps.fold(
         0,
         (previousValue, swap) =>
-            previousValue + swap.myInfo.myAmount.toDouble());
+            previousValue + (swap.myInfo?.myAmount ?? 0));
     return swapFill / order.baseAmount.toDouble();
   }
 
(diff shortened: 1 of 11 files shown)


DeckerSU:
@CharlVS Maybe it would be a good idea to add a separate test (or tests) under `test_units` to deserialize the response of `getMyRecentSwaps` and create different test cases where `my_info` and other parameters are set to `null`. Just to be sure we cover as many of those cases as possible.

What do you think? Is it worth doing?

DeckerSU (approved):
LGTM. 

However I just want to mention that there are plenty of places where the GUI relies on `kdf`. For example, if `events` contains `null`, this expression `List<Map<String, dynamic>>.from(json['events'])` will fail as well, etc.

Also, this isn’t related to these changes, but I noticed that in this build (not sure in which commit it was introduced earlier) on the Windows platform, if a user enters an incorrect wallet password, it doesn’t say "Invalid password" or anything similar. Instead, it responds with "Invalid parameters provided to wallet," which can be confusing. Just letting you know here - I’ll probably create a separate issue for this.

smk762:
If you could please supply the steps to replicate (i.e. spawn a swap with `nulls`), I'm happy to confirm testing. I was unable to see a related issue which might reveal this. 

Equally content to approve based on Decker's tests, as he first reported the problem and would be best placed to confirm it has been resolved.

DeckerSU:
I guess we can merge this into `dev`, since the root of the issue is now resolved and KW can correctly handle `null` in `my_info`.

PR #2875, opened 2025-07-07
update custom token url

smk762:
After some reports of the "custom token import" failing, I did a check of valid urls to confirm our code was correct. 
The only change was for bsc, and I set some others to null where I was unable to get a valid url from known contracts on a network and a variety of "platform" values (e.g. ftm, fantom, ftm20 etc). 

When building this locally to test, I was still unable to import tokens, despite using the contract in the url comments in this PR, so there appears to be a deeper issue, or perhaps this is dead code, and the logic needing a change is in the Flutter DeFi SDK repo. 

image

This error message may be improved to say "No token info found for contract in CoinGecko API", and a log in the console with full url for further debugging.

Curiously, while testing my PR, I was unable to find the expected coingecko api url in either the console or network tabs or my browser.

@CharlVS @takenagain @TazzyMeister I invite you to further investigate and offer a solution.

### DIFF
--- a/lib/bloc/custom_token_import/data/custom_token_import_repository.dart
+++ b/lib/bloc/custom_token_import/data/custom_token_import_repository.dart
@@ -143,7 +143,7 @@ class KdfCustomTokenImportRepository implements ICustomTokenImportRepository {
       final data = jsonDecode(response.body);
       return data;
     } catch (e) {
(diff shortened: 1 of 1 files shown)


CharlVS (approved):
Image
TYSM

@takenagain, please take care of SMK's suggested investigations before the next release.

PR #2914, opened 2025-07-14
fix(ui): wallet list contrast/spacing

CharlVS:


smk762:
> _No description provided._

Well, guess I'll just try and figure it out from the title, the code diff, and visual comparison with prod. 

image

I like it, looks much better, on the front page anyway. Coin page appears to be clinging on to older theme in parts.

image

If that were not the case, I would be tempted to approve based on aesthetic alone - but that might result in me approving what was not actually seeking approval, which can lead to all sorts of malarkey.

Please clarify this in the initial PR comment.

CharlVS:
@smk762 sorry about the lack of description. This only changes the styling for the wallet list in the sign-in dialog.

PR #2934, opened 2025-07-15
fix(ui): position logout dropdown

CharlVS:
## Summary
- recalc dropdown offset when opening so 'Log out' is positioned below the button

## Testing
- `flutter analyze`

------
https://chatgpt.com/codex/tasks/task_e_68768138e74883269c99ed5f09708603


smk762:
btw - a similar displacement of dropdown can occur with the protocol filter in the "add assets" view.

PR #2949, opened 2025-07-17
feat(settings): persist hide zero balance assets

CharlVS:
## Summary
- store the hide zero balance assets toggle in stored settings
- expose the setting in state and bloc events
- update wallet to use the persisted value

## Testing
- `flutter analyze`

------
https://chatgpt.com/codex/tasks/task_e_68792bd9c88883268186b34ef8e1294a


smk762:
- confirmed toggle is functioning as expected
- confirmed list updates as coins with balance are activated (no toggle cycling required)
- confirmed selection persists for same wallet on logout/login.

The selection also persists when I import a second wallet and log in to it, so it appears that the persistence is global, not login specific. Would it be possible to do login based rather than global persistence? If required changes are large, we can delegate this to an issue for future enhancement (e.g. a wallet level toggle persistence system for all in app toggles).

CharlVS:
> Latest commit does not appear to limit toggle state changes to the logged in wallet only. Changes to one login, appear to persist after logout and login with a different wallet.
> 
> This can be moved to issue and approval granted as it is not a blocker and should not delay release.

Thanks. Please do. I will revert the last changes since it may need deeper consideration to get it working smoothly.

PR #2971, opened 2025-07-22
Ride if mounted, die if not

smk762:
An error was sighted in logs:

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: Null check operator used on a null value
#0      State.context (package:flutter/src/widgets/framework.dart:959)
#1      _ConnectWalletButtonState._createPopupDispatcher.<anonymous closure> (package:web_dex/shared/widgets/connect_wallet/connect_wallet_button.dart:113)
#2      _IguanaWalletsManagerState._onLogIn (package:web_dex/views/wallets_manager/widgets/iguana_wallets_manager.dart:318)
#3      _IguanaWalletsManagerState.build.<anonymous closure> (package:web_dex/views/wallets_manager/widgets/iguana_wallets_manager.dart:67)
#4      _BlocListenerBaseState._subscribe.<anonymous closure> (package:flutter_bloc/src/bloc_listener.dart:216)
#5      _rootRunUnary (dart:async/zone.dart:1538)
#6      _CustomZone.runUnary (dart:async/zone.dart:1429)
#7      _CustomZone.runUnaryGuarded (dart:async/zone.dart:1329)
#8      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:381)
#9      _DelayedData.perform (dart:async/stream_impl.dart:573)
#10     _PendingEvents.handleNext (dart:async/stream_impl.dart:678)
#11     _PendingEvents.schedule.<anonymous closure> (dart:async/stream_impl.dart:649)
#12     _rootRun (dart:async/zone.dart:1517)
#13     _CustomZone.run (dart:async/zone.dart:1422)
#14     _CustomZone.runGuarded (dart:async/zone.dart:1321)
#15     _CustomZone.bindCallbackGuarded.<anonymous closure> (dart:async/zone.dart:1362)
#16     _rootRun (dart:async/zone.dart:1525)
#17     _CustomZone.run (dart:async/zone.dart:1422)
#18     _CustomZone.runGuarded (dart:async/zone.dart:1321)
#19     _CustomZone.bindCallbackGuarded.<anonymous closure> (dart:async/zone.dart:1362)
#20     _microtaskLoop (dart:async/schedule_microtask.dart:40)
#21     _startMicrotaskLoop (dart:async/schedule_microtask.dart:49)
```

An agent was consulted:

image

A PR was submitted, and peace was restored.

### DIFF
--- a/lib/shared/widgets/connect_wallet/connect_wallet_button.dart
+++ b/lib/shared/widgets/connect_wallet/connect_wallet_button.dart
@@ -97,8 +97,6 @@ class _ConnectWalletButtonState extends State<ConnectWalletButton> {
   }
 
   PopupDispatcher _createPopupDispatcher() {
-    final TakerBloc takerBloc = context.read<TakerBloc>();
-    final BridgeBloc bridgeBloc = context.read<BridgeBloc>();
 
     return PopupDispatcher(
       borderColor: theme.custom.specificButtonBorderColor,
@@ -108,6 +106,10 @@ class _ConnectWalletButtonState extends State<ConnectWalletButton> {
       popupContent: WalletsManagerWrapper(
         eventType: widget.eventType,
         onSuccess: (_) async {
+          if (!mounted) return;
+
+          final TakerBloc takerBloc = context.read<TakerBloc>();
+          final BridgeBloc bridgeBloc = context.read<BridgeBloc>();
           takerBloc.add(TakerReInit());
           bridgeBloc.add(const BridgeReInit());
           await reInitTradingForms(context);


CharlVS (changes requested):
@smk762 An alternative approach which would solve the root cause would be to store a reference to the blocs before calling `_createPopupDispatcher` since. If we sometimes skip events that need to be fired, we'll sometimes get wonky state. 

Furthermore, instead of having to pass 3 bloc references, it may be cleaner to use a callback function with those pre-async bloc references. Though be careful to avoid falling into "callback hell" where callbacks go multiple layers deep. In that case, it suggests that there's too much logic in the UI and that it should instead be shifted to using bloc events. The UI should contain little/no business logic (we've made great progress, but we're still repaying tech debt on this).

I've attached the three files that require changes. I'd suggest giving the suggestions a shot and then comparing them against the solution I've provided if you have time.

[dex_helpers.dart.txt](https://github.com/user-attachments/files/21365096/dex_helpers.dart.txt)

[connect_wallet_button.dart.txt](https://github.com/user-attachments/files/21365098/connect_wallet_button.dart.txt)

[wallet_main.dart.txt](https://github.com/user-attachments/files/21365099/wallet_main.dart.txt)

**Please test if the changes have any adverse effects to the dex forms. If the code has always thrown an error then fixing it now may actually be detrimental.**

smk762:
Thanks - I collected a few of these during logs review, though some were when exiting a session, which would explain the behavior. I'll circle back to this after the current UX sprint so I can give it proper time and focus

PR #3015, opened 2025-07-29
chore(fiat-onramp): temporarily disable Ramp until API keys are regenerated

takenagain:
## Summary by CodeRabbit

* **Chores**
  * Temporarily disabled "Card Payment" and "Apple Pay" options with the Ramp provider due to unavailable API keys.
  * Removed Ramp as a fiat provider from the list of available options. Ramp-related options will be re-enabled once API keys are available.

### DIFF
--- a/lib/views/fiat/fiat_page.dart
+++ b/lib/views/fiat/fiat_page.dart
@@ -48,7 +48,10 @@ class _FiatPageState extends State<FiatPage> with TickerProviderStateMixin {
   Widget build(BuildContext context) {
     final coinsRepository = RepositoryProvider.of<CoinsRepo>(context);
     final fiatRepository = FiatRepository(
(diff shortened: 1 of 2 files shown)


takenagain:
@coderabbitai review

lib/views/fiat/fiat_page.dart
CharlVS, line 1: Since the `RampFiatProvider()` is disabled, are the changes to `lib/bloc/fiat/models/fiat_payment_method.dart` still necessary? I haven't looked at the full context yet, but I assume the available options are populated dynamically based on the registered providers.

takenagain: The `defaultFiatPaymentMethods` is a placeholder value used when

- there is no valid Fiat amount entered (initial form state), and
- when logging out and resetting the form (before updating providers with a default example fiat value of 10000).

PR #3031, opened 2025-07-31
fix(activation): add assets deactivation state updates and silently activate coins needed for NFT

takenagain:
Fixes the following issues:
- Deactivating a coin in the "Add Assets" page does not reflect on the main wallet page.
- Deactivated NFT coins reappearing after deactivation (e.g. AVAX, FTM, BNB, ETH).
- Possible duplication of coins in the "Add Assets" page when filtering and sorting. This could happen because the legacy usdPrice and address fields were added to the Coin model props for backwards-compatible UI/state updates.
- Filters out NFT_ coins from the private key export list

## Summary by CodeRabbit

* **Refactor**
  * Improved code formatting and readability across several areas of the app.
  * Simplified method signatures and removed unused parameters for cleaner code.
  * Adjusted the order of state updates for more responsive UI during coin removal.
  * Updated activation methods to allow optional addition to wallet metadata and notification control.

* **Style**
  * Unified and streamlined code style for better maintainability and consistency.

### DIFF
--- a/lib/views/settings/widgets/security_settings/security_settings_page.dart
+++ b/lib/views/settings/widgets/security_settings/security_settings_page.dart
@@ -242,8 +242,11 @@ class _SecuritySettingsPageState extends State<SecuritySettingsPage> {
         try {
           // Fetch private keys directly into local UI state
           // This keeps sensitive data in minimal scope
-          _sdkPrivateKeys = (await context.sdk.security.getPrivateKeys())
-            ..removeWhere((asset, _) => excludedAssetList.contains(asset.id));
+          final privateKeys = await context.sdk.security.getPrivateKeys();
+          final filteredPrivateKeyEntries = privateKeys.entries.where(
+            (entry) => !excludedAssetList.contains(entry.key.id),
+          );
+          _sdkPrivateKeys = Map.fromEntries(filteredPrivateKeyEntries);
 
           return true; // Success
         } catch (e) {


takenagain:
@coderabbitai review

lib/views/settings/widgets/security_settings/security_settings_page.dart
```diff
@@ -241,7 +242,8 @@ class _SecuritySettingsPageState extends State<SecuritySettingsPage> {
         try {
           // Fetch private keys directly into local UI state
           // This keeps sensitive data in minimal scope
-          _sdkPrivateKeys = await context.sdk.security.getPrivateKeys();
+          _sdkPrivateKeys = (await context.sdk.security.getPrivateKeys())
+            ..removeWhere((asset, _) => excludedAssetList.contains(asset.id));
```
CharlVS: This is low risk, but rather create a new list instead of modifying the one passed from the SDK (e.g. `.where(...).toList()`. I have implemented measures in the SDK to ensure that we clone before returning any data and/or make it resilient in case it's modified downstream, but it's a good practice nevertheless.

In general (or at least in my opinion), avoid mutating lists that weren't created in the local scope.

CharlVS (changes requested):

lib/views/settings/widgets/security_settings/security_settings_page.dart
takenagain: Done in [5661859](https://github.com/KomodoPlatform/komodo-wallet/pull/3031/commits/566185987494b42bdbca431e3d2cb5c835fba4f6)

PR #3037, opened 2025-08-01
refactor: build.sh, simplify Docker env vars and build command logic

DeckerSU:
- Remove unnecessary environment variables from Docker container
- Keep only GITHUB_API_PUBLIC_READONLY_TOKEN as env var
- Replace hardcoded build logic with conditional credential handling
- Add proper validation for Trello and Cloudflare feedback services
- Use dart-define flags instead of environment variables for build configs
- Add warning messages for incomplete credential sets


DeckerSU:
Rebased against the current `dev` branch and **ready** for merge. Btw, the previous script version contained an error that prevented the build from accessing the real values of the `--dart-define` directives, because they were applied during the assets-fetch stage instead of the final build stage. The current version fixes this bug and has been fully tested in real Jenkins pipelines. Since the submission tests passed, we can confirm that passing define variables to Dart is working.

**p.s.** Please squash this PR into a single commit.

CharlVS (approved):
Thank you! Changes made to remove the test feedback endpoint as discussed.

Issue #2613, opened 2025-05-09
GHA windows desktop builds missing dlls


smk762:
**To Reproduce**
1. Download a recent GHA/CI build for windows desktop
2. Extract and launch
3. See error pop ups

**Screenshots**

![Image](https://github.com/user-attachments/assets/4ad979ea-6af6-4097-a96f-c368510fc205)

![Image](https://github.com/user-attachments/assets/0f72318c-9f24-46b0-95d1-33158012dbb6)

![Image](https://github.com/user-attachments/assets/10c89f66-b8d4-4cc8-acd4-8e049bc30d46)

![Image](https://github.com/user-attachments/assets/ec766aa6-f437-470a-99fa-c024abd5c1b7)

**Expected behavior**
App should launch successfully with required dlls in the app package.

**Found in**
https://github.com/KomodoPlatform/komodo-wallet/actions/runs/14708824239

**Environment:**
- Windows 11 (VM, via VirtualBox 7.18 running on Ubuntu 24)
- Fresh & clean slate, no dev tools etc installed.

smk762:
This issue is still present on fresh windows systems. It was suggested that a user should download the deps, but currently we do not effectively communicate that when the app fails to launch.

Suggestion: On app launch, detect if dlls missing, and if so, show a pop up linking to the required deps install instructions. Would this be possible, or does the missing dll remove the ability to do this also?

smk762:
Comments towards resolution via GPT:

## Problem
Running `komodowallet.exe` results in missing DLL errors:

- `VCRUNTIME140.dll` not found
- `VCRUNTIME140_1.dll` not found
- `MSVCP140.dll` not found
- `flutter_inappwebview_windows_plugin.dll` not found

---

## Cause
1. The `VCRUNTIME140.dll`, `VCRUNTIME140_1.dll`, and `MSVCP140.dll` files are part of the **Microsoft Visual C++ Redistributable**.
2. `flutter_inappwebview_windows_plugin.dll` is part of the application’s Flutter build output, not a system DLL.

---

## Fix

### 1. Install Microsoft Visual C++ Redistributable
Download and install both x64 and x86 packages from Microsoft:

- [vc_redist.x64.exe](https://aka.ms/vs/17/release/vc_redist.x64.exe)  
- [vc_redist.x86.exe](https://aka.ms/vs/17/release/vc_redist.x86.exe)

After installing, restart your system to ensure the new DLLs are recognized.

---

[...]

PR #3063, opened 2025-08-11
fix(windows, linux, macos): enforce single-instance; focus existing window; prevent zombie processes

### LINKED ISSUE #3027: Multiple Komodo Wallet instances can be launched, but only one window is visible
On Windows, it’s possible to start multiple Komodo Wallet processes by clicking the executable repeatedly. However, only the first instance shows a window — additional processes run in the background without any UI, consuming memory and requiring manual termination.

### Steps to Reproduce

1. Launch `komodowallet.exe` and wait for the main window to appear.
2. In Explorer (or via shortcut), double-click `komodowallet.exe` again.
3. Observe that a second Komodo Wallet process starts (e.g. in Task Manager) but no new window is shown.
4. Repeat step 2 several times (5–10 clicks). You will see multiple Komodo Wallet processes (each using \~134 MB RAM) but only one active window.
5. Close the visible Komodo Wallet window.
6. Notice that the “zombie” processes from steps 2–4 remain running without any UI.

### Current Behavior

* Each click of the executable spawns a new background-only process.
* Only one window is ever visible/interactive.
* After closing the visible window, orphaned wallet processes must be killed manually via Task Manager.

### Expected Behavior

* Only one instance of Komodo Wallet should run at any time.
* If the user attempts to launch a second instance:

  * **Option A:** The new process should immediately exit, logging or notifying that an instance is already running.
  * **Option B:** The existing Komodo Wallet window should be brought to the foreground.

### Environment

* **OS:** Windows 11
* **Komodo Wallet version:** \[`dev` branch, `b8bdf68f0`]

### Screenshots

*No error dialogs are shown — the extra instances silently run in the background.*

Image

CharlVS:
# PR Title

fix(windows, linux, macos): enforce single-instance; focus existing window; prevent zombie processes.

Fixes #3027

## Description

- **Summary**: Prevent multiple background-only instances from launching. Ensure subsequent launches activate the existing window instead of spawning "zombie" processes.
- **Why**: On Windows, repeated launches created extra processes without UI, wasting memory and requiring manual termination. Similar safeguards added for other desktop targets to avoid the same class of issue.

### Windows

- Enforce single instance via a named mutex `Global\KomodoWallet_SingleInstanceMutex`.
- If a second process launches, find the existing window by class (`FLUTTER_RUNNER_WIN32_WINDOW`) or title (`komodowallet`), call `ShowWindow(SW_RESTORE)`, `AllowSetForegroundWindow(ASFW_ANY)`, `SetForegroundWindow`, and `BringWindowToTop`, then exit immediately.
- Properly release/close mutex handle on all exit paths.
- File: `windows/runner/main.cpp`

### Linux

- Switch `GtkApplication` to unique mode (remove `G_APPLICATION_NON_UNIQUE`).
- Track `main_window` in app state; on `activate`, present the existing window instead of creating a new one.
- Clear `main_window` on destroy.
- Files: `linux/my_application.cc`, `linux/CMakeLists.txt`

### macOS

- Keep default single-instance behavior; add `applicationShouldHandleReopen` to bring the window to the front when re-launched or re-opened.
- File: `macos/Runner/AppDelegate.swift`

### Not impacted

- Android/iOS are single-instance by OS design; no changes.

### Windows

- Build and run the app.
- Double-click the executable multiple times: only one process should exist; existing window should be focused/restored.
- Close the window: no lingering processes in Task Manager.

### Linux

- Launch the app, then run the binary again from a terminal/launcher: the existing window should be presented; no additional processes should remain after closing.

### macOS

- Launch the app, then re-open from Dock/Spotlight: the existing window should become key and front.

## Risk/Regression

- Low. Desktop-only entrypoints touched; guarded and cleaned up handles appropriately.
- Windows focus may be subject to OS focus-stealing rules; `AllowSetForegroundWindow(ASFW_ANY)` is applied before focusing.

## Files Changed (high-level)

- `windows/runner/main.cpp`
- `linux/my_application.cc`
- `linux/CMakeLists.txt`
- `macos/Runner/AppDelegate.swift`

### DIFF
--- a/linux/my_application.cc
+++ b/linux/my_application.cc
@@ -111,7 +111,7 @@ static void my_application_dispose(GObject* object) {
   g_clear_pointer(&self->dart_entrypoint_arguments, g_strfreev);
 
   if (self->main_window != nullptr) {
-    g_signal_handlers_disconnect_by_func(self->main_window, G_CALLBACK(on_window_destroy), self);
+    g_signal_handlers_disconnect_by_data(self->main_window, self);
     self->main_window = nullptr;
   }
 


CharlVS:
@cursor, is this issue specific to the Windows desktop app, or do Linux and macOS users face a similar problem?

smk762:
I get 
```

/home/smk/GITHUB/KP/komodo-wallet/linux/my_application.cc:113:5: error: no matching function for call to 'g_signal_handlers_disconnect_matched'
/home/smk/GITHUB/KP/komodo-wallet/linux/my_application.cc:133:47: error: 'G_APPLICATION_FLAGS_NONE' is deprecated: Use 'G_APPLICATION_DEFAULT_FLAGS' instead [-Werror,-Wdeprecated-declarations]
Building Linux application...                                           
```
errors on local builds.

Looks like CI is also 
https://github.com/KomodoPlatform/komodo-wallet/actions/runs/16888637218/job/47842558365?pr=3063#step:7:198

smk762:
Using the CI builds in windows still results in https://github.com/KomodoPlatform/komodo-wallet/issues/2613
After manually installing the `Microsoft Visual C++ Redistributable` packages manually, I was still unable to see the app after launch. Checking task manager, it's under "background processes" - shouldn't it be under "apps"?

https://github.com/user-attachments/assets/eb7a4bad-a908-42ab-a6f2-b711e1825314

I'll retry resurrection of physical windows machine again today in case the issues are VM related.

DeckerSU:
Detection of whether the app is already running works incorrectly under Windows. Here is how to check it:

1. Install any file manager, for example, Far Manager from [https://www.farmanager.com/](https://www.farmanager.com/).
2. Start Far Manager, navigate to your Komodo Wallet folder, and run `komodowallet.exe`.
3. You will see a window with Komodo Wallet open.
4. Now open Explorer and navigate to the Komodo Wallet folder.
5. Each time you start `komodowallet.exe` from Explorer, it will create zombie processes with Explorer as the parent.

Below is a screenshot of the process tree. The error is still present. On other platforms, additional checks will be needed, as I have only tested the Windows platform so far.

The working instance of Komodo Wallet:

image

The zombie instances:

image


Maybe it's related to access rights. In my case, Far Manager was likely running with administrative rights, and the process launched from Explorer can't read the mutex. We should probably check not only `GetLastError() == ERROR_ALREADY_EXISTS` but also `ERROR_ACCESS_DENIED`. I'll do some research and let you know how to fix the issue.

DeckerSU:
Windows build seems Ok after [4ef700b](https://github.com/KomodoPlatform/komodo-wallet/pull/3063/commits/4ef700bc49e620f5f9d00956a1beb060d95fb7af), will test other platforms.

smk762:
confirmed functional on physical windows machine. CI builds are failing for other desktop OS though.

DeckerSU:
You merged it too early, without waiting for approval. As a result, the Linux build is now failing with:

```
Building Linux application...                                   
/app/linux/my_application.cc:134:47: error: use of undeclared identifier 'G_APPLICATION_DEFAULT_FLAGS'; did you mean 'FL_APPLICATION_GET_CLASS'?
Build process failed
```

This happens because the `G_APPLICATION_DEFAULT_FLAGS` enumeration value is only available starting from GLib 2.74, released in September 2022.

You can check the GLib version on the build host with:

```bash
dpkg -l 'libglib2.0-0' | awk '/^ii/{print $3}' 
dpkg -l 'libglib2.0-dev' | awk '/^ii/{print $3}' 
```

Our Docker build workflow uses the `docker.io/ubuntu:22.04` image, which includes GLib 2.72.4. That version doesn’t provide `G_APPLICATION_DEFAULT_FLAGS` in the headers, so the build fails.

We need a universal solution that checks the GLib version. I’ll provide it shortly in a separate PR.

DeckerSU:
And on macOS it is still possible to open two windows of KW - for example, one from Finder and another from the console/terminal. Below is a video illustrating this:

https://github.com/user-attachments/assets/6f99c218-8177-4a58-bb60-23001b5af75a

Issue #3137, opened 2025-09-12
.docker/build.sh build fails due to unclean submodules


DeckerSU:
Running `./.docker/build.sh apk release` fails after PR #3125 with the following errors:

```
lib/main.dart:64:53: Error: Member not found: 'SparklineRepository.defaultInstance'.
    final sparklineRepository = SparklineRepository.defaultInstance();
                                                    ^^^^^^^^^^^^^^^
lib/model/coin.dart:112:38: Error: The getter 'displayName' isn't defined for the type 'AssetId'.
 - 'AssetId' is from 'package:komodo_defi_types/src/assets/asset_id.dart' ('sdk/packages/komodo_defi_types/lib/src/assets/asset_id.dart').
Try correcting the name to the name of an existing getter, or defining a getter or field named 'displayName'.
    return _cachedDisplayName ??= id.displayName;
                                     ^^^^^^^^^^^
Target kernel_snapshot_program failed: Exception
```

And the Gradle build aborts:

```
Execution failed for task ':app:compileFlutterBuildRelease'.
> Process 'command '/home/komodo/flutter/bin/flutter'' finished with non-zero exit value 1
```

It looks like `SparklineRepository.defaultInstance()` was removed.

Additionally, `AssetId.displayName` is no longer available.

DeckerSU:
```
  factory SparklineRepository.defaultInstance() {
    return SparklineRepository([
      BinanceRepository(binanceProvider: const BinanceProvider()),
      CoinPaprikaRepository(coinPaprikaProvider: CoinPaprikaProvider()),
      CoinGeckoRepository(coinGeckoProvider: CoinGeckoCexProvider()),
    ], selectionStrategy: DefaultRepositorySelectionStrategy());
  }
```
Seems introduced only in SDK - `3f503d2368c6ddd7a03bb15b587c3f27efa845c5` (`sparkline_repository.dart`), but the app in latest dev branch seems still use `f22ca695471a52f3409726d01116fc811cee6751`.

```
git submodule status
+f22ca695471a52f3409726d01116fc811cee6751 sdk (komodo_cex_market_data-v0.0.3+1-2-gf22ca69)
```
That's why it fails.

DeckerSU:
After further investigation and log analysis, I came to the conclusion that the issue might be in the `.docker/build.sh` script. It’s possible that the `sdk` submodules could be left unupdated, causing the build to fail. Still investigating...
```
[...]

PR #3139, opened 2025-09-12
ci(docker): clean submodules before update to fix build errors

DeckerSU:
Added clean submodules step before `git submodule update` in the build script to ensure submodules are always in a clean state before checkout. This prevents local changes inside `sdk` (e.g. changed build_config.json from assets transformer) from blocking the update step.

Fixes: https://github.com/KomodoPlatform/komodo-wallet/issues/3137


smk762 (approved):
Good call, thanks

CharlVS (approved):

Issue #2883, opened 2025-07-08
Trezor passphrase input can not be unhidden


smk762:
> ![Image](https://github.com/user-attachments/assets/f49c4a73-9202-48eb-8089-b51e2d75250e)
> 
> We should also include a view/hide icon to this field. 

 _Originally posted by @smk762 in [#2766](https://github.com/KomodoPlatform/komodo-wallet/issues/2766#issuecomment-3031609579)_

smk762:
Issue still present in 0.9.3 RC

PR #3214, opened 2025-10-23
Trezor visibility 

smk762:
Closes https://github.com/KomodoPlatform/komodo-wallet/issues/2883

image

To test:
With a Trezor 5 Safe or Trezor Model One, initiate trezor login
- [ ] Confirm the passphrase input has a functioning visibility toggle.
- [ ] Confirm the "connect hardware wallet" option is not visible on mobile apps


dimxy:
I tried to enter pin with Trezor One.
indeed, when visibility is on, the input PIN entry shows the positions from the 3 x 3 numpad matched to the scrambled pin on device. I think showing them may be confusing and it's better to show just "****" instead.

smk762:
> I tried to enter pin with Trezor One. indeed, when visibility is on, the input PIN entry shows the positions from the 3 x 3 numpad matched to the scrambled pin on device. I think showing them may be confusing and it's better to show just "****" instead.

Agree - removed. It's likely only to lead to confusion.

dimxy (approved):

Issue #3275, opened 2025-10-28
Temporary app freeze on `unban_pubkeys` in android


smk762:
**To Reproduce**
1. Log in 
2. Go to settings > security
3. Click the unban pubkeys button
4. See screen go white

**Expected behavior**
A simple pop up to ensure me that pubkeys have indeed been, unbanned.

**Found in**
https://github.com/KomodoPlatform/komodo-wallet/pull/3253

**Environment:**
 - Platform: Mobile (unable to replicate in web or linux)
 - Device: Samsung S21

**Additional context**
Logs below:

```

2025-10-28: 6:17:20.468T+:0:05:02.037875 [LOG] {message: Uncaught exception: Null check operator used on a null value.
#0      _ResponsiveBody._buildTrailingWidget (package:web_dex/views/settings/widgets/security_settings/security_action_plate.dart:196)
#1      _ResponsiveBody._buildColumnLayout (package:web_dex/views/settings/widgets/security_settings/security_action_plate.dart:145)
#2      _ResponsiveBody.build (package:web_dex/views/settings/widgets/security_settings/security_action_plate.dart:116)
#3      StatelessElement.build (package:flutter/src/widgets/framework.dart:5791)
#4      ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:5723)
#5      Element.rebuild (package:flutter/src/widgets/framework.dart:5435)
#6      StatelessElement.update (package:flutter/src/widgets/framework.dart:5797)
#7      Element.updateChild (package:flutter/src/widgets/framework.dart:3982)
#8      ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:5747)
#9      Element.rebuild (package:flutter/src/widgets/framework.dart:5435)
#10     StatelessElement.update (package:flutter/src/widgets/framework.dart:5797)
#11     Element.updateChild (package:flutter/src/widgets/framework.dart:3982)
#12     ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:5747)
#13     StatefulElement.performRebuild (package:flutter/src/widgets/framework.dart:5884)
#14     Element.rebuild (package:flutter/src/widgets/framework.dart:5435)
#15     StatefulElement.update (package:flutter/src/widgets/framework.dart:5909)
#16     Element.updateChild (package:flutter/src/widgets/framework.dart:3982)
#17     ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:5747)
[...]

Issue #3026, opened 2025-07-30
`Null check operator used on a null value` on desktop login


smk762:
**To Reproduce**
- Login

**Screenshots**

https://github.com/user-attachments/assets/f82a8783-8756-4a8a-9daf-93d9c0f65e6f

**Logs**
```
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: Null check operator used on a null value
#0      State.context (package:flutter/src/widgets/framework.dart:959)
#1      _ConnectWalletButtonState._createPopupDispatcher.<anonymous closure> (package:web_dex/shared/widgets/connect_wallet/connect_wallet_button.dart:113)
#2      _IguanaWalletsManagerState._onLogIn (package:web_dex/views/wallets_manager/widgets/iguana_wallets_manager.dart:321)
#3      _IguanaWalletsManagerState.build.<anonymous closure> (package:web_dex/views/wallets_manager/widgets/iguana_wallets_manager.dart:68)
#4      _BlocListenerBaseState._subscribe.<anonymous closure> (package:flutter_bloc/src/bloc_listener.dart:216)
#5      _rootRunUnary (dart:async/zone.dart:1538)
#6      _CustomZone.runUnary (dart:async/zone.dart:1429)
#7      _CustomZone.runUnaryGuarded (dart:async/zone.dart:1329)
#8      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:381)
#9      _DelayedData.perform (dart:async/stream_impl.dart:573)
#10     _PendingEvents.handleNext (dart:async/stream_impl.dart:678)
#11     _PendingEvents.schedule.<anonymous closure> (dart:async/stream_impl.dart:649)
#12     _rootRun (dart:async/zone.dart:1517)
#13     _CustomZone.run (dart:async/zone.dart:1422)
#14     _CustomZone.runGuarded (dart:async/zone.dart:1321)
#15     _CustomZone.bindCallbackGuarded.<anonymous closure> (dart:async/zone.dart:1362)
#16     _rootRun (dart:async/zone.dart:1525)
#17     _CustomZone.run (dart:async/zone.dart:1422)
#18     _CustomZone.runGuarded (dart:async/zone.dart:1321)
#19     _CustomZone.bindCallbackGuarded.<anonymous closure> (dart:async/zone.dart:1362)
#20     _microtaskLoop (dart:async/schedule_microtask.dart:40)
[...]

PR #3276, opened 2025-10-28
fix(settings): prevent null-assert on Android by using disabled trailing button for unban_pubkeys

smk762:
closes https://github.com/KomodoPlatform/komodo-wallet/issues/3275

To test:
- login on android app
- go to settings > security
- spam press the unban pubkeys button
- [ ] confirm no app freeze or white screen

----

### Summary
- **What**: Replace default button wiring with a custom trailing button in `UnbanPubkeysPlate` so the button can be disabled during RPC.
- **Why**: Prevents a “Null check operator used on a null value” in `SecurityActionPlate` during loading.
- **Scope**: Single-file change, no shared component edits.

### Context
- Issue: temporary white screen on Android when tapping “Unban pubkeys” due to a null-assert in `_buildTrailingWidget`.
- Reference: [#3275](https://github.com/KomodoPlatform/komodo-wallet/issues/3275) (related: #3026).
- This PR fixes: Fixes [#3275](https://github.com/KomodoPlatform/komodo-wallet/issues/3275)

### Root cause
- `SecurityActionPlate` assumes `onActionPressed` is non-null when `trailingWidget == null` and force-unwraps it.
- `UnbanPubkeysPlate` sets `onActionPressed: null` while loading to disable the button, violating that assumption and causing the crash.

### Changes
- In `UnbanPubkeysPlate`, provide a `trailingWidget` using `UiPrimaryButton`:
  - Disabled while loading (`onPressed: null`).
  - Keeps existing width/height and text styling parity with the plate’s default button.

### Modified files
- `lib/views/settings/widgets/security_settings/unban_pubkeys_plate.dart`

### Behavior
- Before: Intermittent white screen on Android after tapping “Unban pubkeys”; app recovers after app-switching.
- After: Button shows disabled/loading state; no crash; dialog/snackbar shows results as expected.

### Manual test plan
- Login → Settings → Security → Tap “Unban pubkeys”.
- Verify:
  - Button disables and shows loading text.
  - On completion:
    - Dialog with results if any banned keys existed, or
    - Snackbar indicating no banned pubkeys.
- Platforms:
  - Android (e.g., Samsung S21 as per report).
  - Optional sanity on Web/Linux to ensure unchanged behavior.

### Risk
- Low. Change is isolated to `UnbanPubkeysPlate`. Uses existing `UiPrimaryButton` disabled handling.

### DIFF
--- a/lib/analytics/widgets/analytics_lifecycle_handler.dart
+++ b/lib/analytics/widgets/analytics_lifecycle_handler.dart
@@ -69,10 +69,18 @@ class _AnalyticsLifecycleHandlerState extends State<AnalyticsLifecycleHandler>
   void didChangeAppLifecycleState(AppLifecycleState state) {
     super.didChangeAppLifecycleState(state);
 
+    log('AnalyticsLifecycleHandler: App lifecycle state changed to $state');
+
     // Log app opened event when app is resumed (but not on initial open)
     if (state == AppLifecycleState.resumed && _hasLoggedInitialOpen) {
+      log('AnalyticsLifecycleHandler: App resumed, triggering health check after backoff');
       _logAppOpenedEvent();
-      _checkAuthStatus();
+      // Add 150ms backoff before health check to avoid race where native status
+      // reports "running" but HTTP listener hasn't bound yet after iOS backgrounding
+      Future.delayed(const Duration(milliseconds: 150), () {
+        log('AnalyticsLifecycleHandler: Backoff complete, checking auth status');
+        _checkAuthStatus();
+      });
     }
   }
 
--- a/lib/bloc/dex_tab_bar/dex_tab_bar_bloc.dart
+++ b/lib/bloc/dex_tab_bar/dex_tab_bar_bloc.dart
@@ -93,7 +93,7 @@ class DexTabBarBloc extends Bloc<DexTabBarEvent, DexTabBarState> {
 
   FutureOr<void> _onTabChanged(TabChanged event, Emitter<DexTabBarState> emit) {
     // Validate tabIndex to prevent out-of-bounds access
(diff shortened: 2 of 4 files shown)


lib/views/settings/widgets/security_settings/unban_pubkeys_plate.dart
```diff
@@ -36,12 +39,18 @@ class _UnbanPubkeysPlateState extends State<UnbanPubkeysPlate> {
...
+              onPressed:
+                  state.isUnbanningPubkeys ? null : () => _handleUnbanPressed(context),
+              width: isMobile ? double.infinity : 210.0,
+              height: isMobile ? 52.0 : 40.0,
+              text: state.isUnbanningPubkeys
+                  ? '${LocaleKeys.unbanPubkeys.tr()}...'
+                  : LocaleKeys.unbanPubkeys.tr(),
+              textStyle: Theme.of(context).textTheme.labelLarge?.copyWith(
+                color: theme.custom.defaultGradientButtonTextColor,
+              ),
```
CharlVS, line 52: Is overriding the `textStyle` necessary here? In general, we want to avoid overriding theme values or setting on specific widgets as much as possible since it makes life much easier when we want to change the app's styling. Flutter's theming system automagically handles things like contrast and "looking nice".

smk762: image

Seems this is also applied to `plate_seed_backup.dart` & `security_action_plate.dart` in the same folder.
Should I remove the `textStyle` bit for all 3?

CharlVS: Since it maintains consistency with the other (even tho no ideal) usages, no changes needed.

CharlVS (approved):

PR #3300, opened 2025-10-29
fix(trading-bot): use max_maker_vol for spendable balance

takenagain:
Replaces references to spendable balance with `max_maker_vol`. Spendable balance is the sum total of all funded addresses in HD mode, which results in base volume too low errors when the trade pre-image check only uses the swap address balance.

## Steps to reproduce

1. Create or import an HD wallet
2. Navigate to a coin details page and create 2 additional addresses 
3. Fund all 3 addresses of the coin
4. Enable trading bot in `Settings -> General -> Enable Trading Bot`
5. Navigate to the Trading Bot page, and 
6. Observe that the tradeable amount is the total of the funded addresses, rather than the maximum maker volume of the swap address.

## Before 

image

image

## After 

image

<details><summary>Before fiat amount UI fix</summary>
image
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,102 +1,224 @@
 # Komodo Wallet v0.9.3 Release Notes
 
-This release sharpens sign-in convenience, analytics coverage, and operational resilience ahead of the 0.9.3 cut. Highlights include one-click remember-me login, a dual analytics pipeline with Matomo support, a hardened feedback workflow, and improved compliance guardrails across platforms.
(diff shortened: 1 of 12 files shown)


cipig:
> This is incongruent with the surrounding numbers.

is the param `min_volume` passed to `setprice` call when posting the order?
if so, then it needs to be > 0 and > "default min_volume" of the coin
if there is no intention to set an explicit min_volume, then better don't pass min_volume in setprice, but let kdf use the default one (10x txfee from coins file)

there is this call: https://komodoplatform.com/en/docs/komodo-defi-framework/api/legacy/min_trading_vol/
and it returns this for KMD:
```
   "result" : {
      "coin" : "KMD",
      "min_trading_vol" : "0.0001",
      "min_trading_vol_fraction" : {
         "denom" : "10000",
         "numer" : "1"
      },
```
10000 = 1000 * 10 ... 1000 is from `"txfee": 1000` in coins file for KMD

smk762 (approved):
LGTM, thanks!

Issue #3307, opened 2025-10-30
ZHTLC transactions should not need to be generated twice


smk762:
This is exactly the same problem as present in https://github.com/KomodoPlatform/komodo-wallet/issues/3196

It's actually universal, but goes unnoticed for other coins as the process is too fast to notice.

smk762:
closing as complete

Issue #3196, opened 2025-10-17
Trezor withdraws should not require two signatures


smk762:
**To Reproduce**
1. Login with trezor
2. Create a withdraw
3. Sign it on device
4. click confirm in gui
5. See it want another signing.

**Screenshots**

https://github.com/user-attachments/assets/a179fd3d-d82b-4635-bba0-8bbc291982bc

**Expected behavior**
Sign once, then broadcast.

**Found in**
0.9.3 RC

**Environment:**
 - Devices: Trezor 5 Safe & Trezor Model One
 
**Additional context**
Sound on.

PR #3309, opened 2025-10-30
use withdraw preview signed hex for broadcast

smk762:
Closes https://github.com/KomodoPlatform/komodo-wallet/issues/3307 & https://github.com/KomodoPlatform/komodo-wallet/issues/3196

To test:
- [ ] Make a withdrawal with a trezor. You should only need to sign the hex once.
- [ ] Make a withdrawal with of ARRR. You should only need to generate the hex once.

----------

### Summary
Avoids calling the withdraw RPC twice by reusing the transaction preview (`tx_hex`) for broadcast on confirmation. This eliminates a second device-sign prompt for hardware wallets (Trezor) and removes unnecessary duplicate transaction generation on slow chains (e.g., ZHTLC).

### Linked Issues
- Trezor withdraws should not require two signatures [(#3196)](https://github.com/KomodoPlatform/komodo-wallet/issues/3196)
- ZHTLC transactions should not need to be generated twice [(#3307)](https://github.com/KomodoPlatform/komodo-wallet/issues/3307)

### Problem
Current flow generates the transaction twice:
1) First withdraw call for preview (user reviews fees/details)
2) Second withdraw call on confirm (broadcast)

For hardware wallets, this results in two device confirmations for the same transaction. For chains with slower generation (ZHTLC), the duplicate work is visible and wasteful.

### Approach
Perform a single withdraw (preview) to generate the transaction and cache it in local state. On confirm, broadcast the cached `tx_hex` using `send_raw_transaction` instead of calling withdraw again.

### Implementation Details
- Keep the first `WithdrawalPreview` in `WithdrawFormState.preview` when the confirmation screen is shown.
- On Confirm:
  - Call `send_raw_transaction` with `preview.txHex` and `preview.coin`.
  - Build `WithdrawalResult` using the preview data and the broadcast `txHash`.
- Clear cached preview on Cancel/Back (via reset) or successful broadcast.
- Do not re-trigger Trezor signing on confirm; the only user interaction with the device happens during preview/initial signing.

### Files Touched (high level)
- `lib/bloc/withdraw_form/withdraw_form_bloc.dart`
  - Inject `Mm2Api` into the bloc.
  - Replace second withdraw call with `send_raw_transaction` using cached `preview.txHex`.
  - Clear preview on success to avoid stale state.
- `lib/views/wallet/coin_details/withdraw_form/withdraw_form.dart`
  - Provide `Mm2Api` to `WithdrawFormBloc` constructor.

### User Impact
- Trezor: single device-sign prompt, then immediate broadcast on Confirm (no second prompt).
- ZHTLC and other slow chains: no second generation delay; Confirm broadcasts immediately.

### Testing Steps
1) Login with Trezor.
2) Create a withdraw and press Preview.
   - Expect device to prompt once for signing (if required by the chain).
   - Preview screen shows amounts and network fee.
3) Press Confirm.
   - Expect immediate broadcast using the cached `tx_hex`.
   - No second Trezor prompt.
4) Cancel/back from confirm screen.
   - Preview cache is cleared (no leak into next attempts).
5) Repeat on a ZHTLC asset.
   - Confirm should not trigger a second generation; broadcast is immediate.

### Edge Cases
- If `preview` is unexpectedly null on Confirm, an error is surfaced and the transaction fails gracefully.
- If `send_raw_transaction` returns an error, the failure is shown on the failed screen with details.

### Risks and Mitigations
- Risk: Preview cache not cleared on all exits.
  - Mitigation: Cache cleared on success and via existing reset on Cancel/Back.
- Risk: Divergence from legacy paths.
  - Mitigation: The approach mirrors existing KMD rewards claim flow that uses `send_raw_transaction` after constructing `tx_hex`.

### Checklist
- [x] Reuse preview `tx_hex` for broadcast on Confirm
- [x] Remove second device-sign prompt for Trezor
- [x] Clear cached preview on cancel/back and on success
- [x] Minimal surface-area changes (bloc + view wiring only)

### DIFF
--- a/lib/bloc/auth_bloc/auth_bloc.dart
+++ b/lib/bloc/auth_bloc/auth_bloc.dart
@@ -98,6 +98,10 @@ class AuthBloc extends Bloc<AuthBlocEvent, AuthBlocState> with TrezorAuthMixin {
       // Log and continue to clear local auth state so UI can recover.
       _log.shout('Error during sign out, proceeding to reset state', e, s);
     } finally {
+      // Explicitly disconnect SSE on sign-out
+      _log.info('User signed out, disconnecting SSE...');
+      _kdfSdk.streaming.disconnect();
+      
       await _authChangesSubscription?.cancel();
       emit(AuthBlocState.initial());
     }
@@ -139,6 +143,11 @@ class AuthBloc extends Bloc<AuthBlocEvent, AuthBlocState> with TrezorAuthMixin {
 
       _log.info('Successfully logged in to wallet');
       emit(AuthBlocState.loggedIn(currentUser));
+      
+      // Explicitly connect SSE after successful login
+      _log.info('User authenticated, connecting SSE for streaming...');
+      _kdfSdk.streaming.connectIfNeeded();
+      
       _listenToAuthStateChanges();
     } catch (e, s) {
       if (e is AuthException) {
@@ -453,6 +462,17 @@ class AuthBloc extends Bloc<AuthBlocEvent, AuthBlocState> with TrezorAuthMixin {
           ? AuthorizeMode.logIn
           : AuthorizeMode.noLogin;
       add(AuthModeChanged(mode: event, currentUser: user));
+      
+      // Tie SSE connection lifecycle to authentication state
+      if (user != null) {
+        // User authenticated - connect SSE for balance/tx history streaming
+        _log.info('User authenticated, connecting SSE for streaming...');
+        _kdfSdk.streaming.connectIfNeeded();
+      } else {
+        // User signed out - disconnect SSE to clean up resources
+        _log.info('User signed out, disconnecting SSE...');
+        _kdfSdk.streaming.disconnect();
+      }
     });
   }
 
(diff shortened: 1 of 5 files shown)


ca333 (approved):

lib/bloc/withdraw_form/withdraw_form_bloc.dart
```diff
@@ -21,12 +23,15 @@ import 'package:decimal/decimal.dart';
 class WithdrawFormBloc extends Bloc<WithdrawFormEvent, WithdrawFormState> {
   final KomodoDefiSdk _sdk;
   final WalletType? _walletType;
+  final Mm2Api _mm2Api;
 
   WithdrawFormBloc({
     required Asset asset,
     required KomodoDefiSdk sdk,
```
CharlVS, line 30: @smk762 Thank you for this PR. Btw, for the future, the SDK has 1:1 type-safe mapping for most RPCs, which is accessible through the SDK, so passing `_mm2Api` isn't necessary here. e.g. `_sdk.rpc.withdraw.withdraw` or `_sdk.rpc.withdraw.sendRawTransaction`.

But in general, we want to avoid having to use RPCs instead of the abstracted/managed SDK features so that devs don't need to learn and understand KDF's API. With the SDK's withdrawal feature now unused, it suggests it doesn't meet our needs, and therefore won't meet the needs of other developers.

PR #3329, opened 2025-10-31
chore(settings): add logging and avoid silent skipping

takenagain:
Adds logging for visibility of any errors or skipped configs


smk762 (approved):
Thanks for the extras

PR #3327, opened 2025-10-31
fix(tx-history): only consider empty timestamp and confs as unconfirmed

takenagain:
Only filters a transaction to the top as unconfirmed if both the timestamp and confirmations are 0 (timestamp = epoch).

## Steps to reproduce

1. Import a wallet with a coin with transaction history on the longer side (20+ transactions going back months to years).
2. Activate a coin with transaction history, if not already active
3. Open the coin details page quickly after activation
4. Observe the transactions trickling into transaction history and confirm that they are sorted by timestamp
5. Create a transaction by receiving to or sending funds
6. Observe the transaction history and confirm that the newest transaction appears at the top with a Date of "Now"
7. Wait for the transaction to be confirmed and observe the date change from "Now" to a valid timestamp for the new transaction while in the coin details page

PR #3338, opened 2025-11-02
fix(transaction-history): migrate from confirmations to timestamp-based sorting

DeckerSU:
This reverts the changes from https://github.com/KomodoPlatform/komodo-wallet/commit/9900372bde0aa9f1ca957a3a1ba7ed83ced36f26 that attempted to show unconfirmed transactions first. The previous approach caused transactions to appear in a chaotic order during synchronization, creating confusion until all transactions are fully loaded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,102 +1,224 @@
 # Komodo Wallet v0.9.3 Release Notes
 
-This release sharpens sign-in convenience, analytics coverage, and operational resilience ahead of the 0.9.3 cut. Highlights include one-click remember-me login, a dual analytics pipeline with Matomo support, a hardened feedback workflow, and improved compliance guardrails across platforms.
+This release delivers significant performance improvements, enhanced analytics capabilities, and a comprehensive overhaul of authentication and wallet management. Key highlights include real-time portfolio streaming, a dual analytics pipeline with persistent queueing, one-click sign-in, Z-HTLC support, and extensive optimisations that reduce RPC usage while improving responsiveness across all platforms.
 
 ## 🚀 New Features
 
-- **One-Click Remember Me Sign-In** ([@CharlVS], #3041) - Adds a quick login toggle that remembers hashed wallet metadata, auto-detects password manager autofill, and lets you resume the last wallet in one tap across desktop and mobile.
-- **Feedback Portal Overhaul** ([@CharlVS], #3017) - Rebuilds the in-app feedback flow with provider plugins, optional contact opt-out, log attachments capped under 10 MB, and screenshot scrubbing for sensitive dialogs.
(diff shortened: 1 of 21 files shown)


smk762:
Possible duplication of https://github.com/KomodoPlatform/komodo-wallet/pull/3327
@takenagain pls coordinate with @DeckerSU to resolve these PRs

DeckerSU:
> Possible duplication of #3327 @takenagain pls coordinate with @DeckerSU to resolve these PRs

Yes, I guess it's a duplication. My suggestion is to let the UI team decide which option fits the codebase better. From my point of view, reverting the unconfirmed transactions logic seems a bit cleaner than adding new conditions for confirmations. But that’s just my humble opinion. The final decision on which one should be closed and which should be merged is up to the UI team.

lib/bloc/transaction_history/transaction_history_bloc.dart
```diff
@@ -249,24 +249,13 @@ class TransactionHistoryBloc
...
-    final rightIsUnconfirmed = right.confirmations == 0;
-
-    if (leftIsUnconfirmed != rightIsUnconfirmed) {
-      return leftIsUnconfirmed ? -1 : 1;
+    final now = DateTime.now();
+    if (right.timestamp == now) {
+      return 1;
+    } else if (left.timestamp == now) {
+      return -1;
     }
```
takenagain: Could we change the unconfirmed transaction check to use the Unix epoch instead of DateTime.now(), to bring unconfirmed transactions to the top of the list? 

The 0 timestamps of unconfirmed transactions are parsed using `fromMillisecondsSinceEpoch`, resulting in Unix epoch Date Time values for unconfirmed transactions (in [transaction.dart](https://github.com/KomodoPlatform/komodo-defi-sdk-flutter/blob/dev/packages/komodo_defi_types/lib/src/transactions/transaction.dart#L147)). This is similar to the conditions already present in [formatters.dart](https://github.com/KomodoPlatform/komodo-wallet/blob/5da3aaf9d7ee9fc3681cf270f9a1e999ca208947/lib/shared/utils/formatters.dart#L357). 

```suggestion
    if (right.timestamp == DateTime.fromMillisecondsSinceEpoch(0)) {
      return 1;
    } else if (left.timestamp == DateTime.fromMillisecondsSinceEpoch(0)) {
      return -1;
    }
```

DeckerSU: > Could we change the unconfirmed transaction check to use the Unix epoch instead of DateTime.now(), to bring unconfirmed transactions to the top of the list?

Done in https://github.com/KomodoPlatform/komodo-wallet/pull/3338/commits/e2d8273d31f33c84af64cd0356d55d239a609554 . However, for coins with a large number of transactions in the list, this will not immediately bring the unconfirmed transaction to the top. This behavior is already described here: [https://github.com/KomodoPlatform/komodo-wallet/issues/3342](https://github.com/KomodoPlatform/komodo-wallet/issues/3342)

Issue #2891, opened 2025-07-09
Show swap uuid and allow export of swap data easily


gcharang:
<img width="2614" height="1606" alt="Image" src="https://github.com/user-attachments/assets/983e3db4-d727-40ad-91fb-31047cbe40da" />

In this screen, display the swap uuid and also allow easy copying of it with a button. also provide a export swap data button that will download the swap's data in a json on click

also, allow download of the whole swap data from settings on button click instead of just showing it in the app

smk762:
- [ ] display swap uuid in swap details modal, with copy button
- [ ] Add export button for swap data json in swap detail modal
- [ ] Add export button for **all** swap data json (via settings > show swap data)

gcharang:
updated to P0 as sharing a single swap's data is impossible right now

gcharang:
implemented in 0.9.3 but need to beautify. there are 2 swap id copy lines, remove one of them. Move the "export swap data" button on the swap page to the top, just below the swap uuid line

also add the export swap data button and swap uuid to each swap row in the history/"in progress" tables

fix translation strings if necessary

gcharang:
the exported swap data file for a single file has the wrong swap id in  file name and ends with .txt on ios. make sure the file gets the correct name and ends with .json

CharlVS:
Closing as resolved in PR #3348 (merged 2025-11-23).

PR #3348, opened 2025-11-03
fix: uuid copy / swap export / translations

smk762:
Fixes https://github.com/KomodoPlatform/komodo-wallet/issues/2891 additional comments.
Adding export to rows deferred until design can be determined.

- exports as prettified json
- exports as json file (not`txt`)
- moves export button into pair header block
- removes duplicated swap uuid copy line

image


gcharang:
> Adding export to rows deferred until design can be determined.

it can just be a download icon beside the swap id. it can look like:

Screenshot 2025-11-03 190832

smk762:
> > Adding export to rows deferred until design can be determined.
> 
> it can just be a download icon beside the swap id. it can look like:
> 

My main concern is how adding something so wide will work on mobile. Its a bit tricker question than the rest of the requested changes, so I did the easy, and kicked the can down the road for the difficult. As there now exists a vector to export which is a click away from the list, I figured that the additional button was a convenience rather than critical requirement, and a worthy candidate for triage to be deferred pending multi-platform design.

gcharang:
@smk762 got it. can only check in ios after push to testflight. but on web, LGTM

gcharang (approved):

PR #3352, opened 2025-11-03
fix(tx-history): timestamp update without navigation

takenagain:
Small out-of-scope fix for #3338 that I couldn't add as a suggestion comment earlier. This adds the `timestamp` field to the list of fields updated on each SSE event, which will update the details of unconfirmed transactions if a user stays on the coin details page, instead of remaining labelled as "Now".


smk762:
UTXO Updates are much more rapid :heavy_check_mark: :fire: 
One flaw - I noticed the flip from "unconfirmed" to "now" was premature. Checking the block explorer revealed that the transaction was, in fact, still unconfirmed.

EVM updates still appear to be non functional (or slow enough to appear so). 8 mins, 300+ confs, no update while on same page. Navigation did show the update. **New Issue at: https://github.com/KomodoPlatform/komodo-wallet/issues/3364**

image

IRIS was fast, OSMO not (this may be due to 3rd part nodes?) - but until navigation, the value sent for IRIS was shown as zero. **New Issue at https://github.com/KomodoPlatform/komodo-wallet/issues/3360**

image


ARRR appeared soon after the block was confirmed. This might incur a delay or 1 minute on average, but due to the nature of ARRR, I assume the mempool is not transparent enough to improve on this.

DeckerSU:
I'll approve it today - I just want to run some internal tests first. I'm also monitoring SSE to see how quickly KDF pushes new transactions right after they're created.

CharlVS:
@ smk762, please check if you see the EVM balance streaming events in your logs. KDF doesn't support EVM tx history streaming, but it does support balance streaming, so we should still manually update the TX history when a balance change event occurs.

Could you please open an issue for it?

smk762:
> @ smk762, please check if you see the EVM balance streaming events in your logs. KDF doesn't support EVM tx history streaming, but it does support balance streaming, so we should still manually update the TX history when a balance change event occurs.
> 
> Could you please open an issue for it?

Done : https://github.com/KomodoPlatform/komodo-wallet/issues/3364

smk762 (approved):

PR #3372, opened 2025-11-05
feat: add photo library usage description to Info.plist for QR code i…

DeckerSU:
…mport functionality

Without these changes, we will get the following error when uploading to the App Store:

image


CharlVS:
Can't test this (at least on my iOS version) since it's only applicable when submitting for publication.

CharlVS:
@DeckerSU have you added all necessary perms from https://github.com/miguelpruivo/flutter_file_picker/wiki/Setup#--ios ? As a preventative measure for issues not yet discovered.

DeckerSU:
> As a preventative measure for issues not yet discovered.

No, I’m not a big fan of adding all permissions preemptively. I’d prefer to add any additional ones only if testing shows they’re actually needed for the file picker to work properly. I think it’s better to focus on testing all the areas where file dialogs are used instead.

PR #3374, opened 2025-11-05
fix(ios): fix download logs dialog not opening on iOS

DeckerSU:
Fix issue where the "Download logs" button in Settings was not opening the share dialog on iOS devices.

Changes:

- Migrate from deprecated Share.shareXFiles() to SharePlus.instance.share() API
- Add sharePositionOrigin parameter using scaffoldKey.currentContext for proper iPad support
- Explicitly set mimeType and name parameters for XFile instances
- Add break statements after each case in switch statement (required by Dart)

This ensures the share sheet dialog appears correctly on both iPhone and iPad devices when downloading log files.


CharlVS (approved):
Working on macOS. Thanks!

PR #3398, opened 2025-11-17
use latest KDF commit on `sia-rc-cleanup`

smk762:
PR for CI builds to test SIA integration.  
Please add any problems or missing functionality to https://github.com/KomodoPlatform/komodo-wallet/issues/3399

Confirmed the following issues have been resolved:
- incomplete transaction history display
- No SIA option in filter
- Bypass SIA privkey export error gracefully
- SIA withdraw method and raw transaction broadcast tweaks.
- KDF side maker bug causing swap fails
- KDF side bestorders SIA display bug

---

> [!NOTE]
> Moves coins manager filtering to SDK CoinSubClass, executes withdrawals via SDK progress API, and adds Sia plus tx_json support in requests/models.
> 
> - **Coins Manager (UI/State)**
>   - Switch filter model from legacy `CoinType` to SDK `CoinSubClass` across events, state, bloc logic, and UI components.
>   - Build filter options dynamically from coins’ `id.subClass`; filtering now matches `coin.id.subClass`.
> - **Withdrawals**
>   - Replace direct MM2 `send_raw_transaction` broadcast with `_sdk.withdrawals.executeWithdrawal(...)` progress stream; remove unused `Mm2Api` field.
> - **MM2 API & Models**
>   - `SendRawTransactionRequest`: allow either `tx_hex` or `tx_json` (both optional but one required).
>   - `WithdrawDetails`: accept optional `txHex`/`txJson` with assertion; parsing updated accordingly.
> - **Protocol/Coin Types**
>   - Add Sia mapping between `CoinSubClass` and legacy `CoinType`, including name/color and history/explorer handling.
> - **KMD Rewards**
>   - Validate `txHex` presence before broadcasting.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 57ece5774aed05751092d682bcb47bb8683ad0b5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #3405, opened 2025-11-19
fix single tx history display for SIA coins

smk762:
Fixes bug where only a single tx entry would appear in SIA coin page tx history.

- Added a _txKey helper in transaction_history_bloc.dart that falls back to txHash/id when internalId is empty, ensuring Sia transactions receive unique keys even though MM2 returns "" for internal_id.
- Updated all in-memory maps (byId, _firstSeenAtById, _processedTxIds) to use _txKey(...) so batches no longer overwrite each other when internal_id is blank, letting every SC transaction display.

To test:
- login to a wallet which has more than one SC transaction in its history
- activate SC
- go to SC page
- confirm tx history is complete, and not redacted after a single entry

### DIFF
--- a/lib/bloc/transaction_history/transaction_history_bloc.dart
+++ b/lib/bloc/transaction_history/transaction_history_bloc.dart
@@ -87,22 +87,24 @@ class TransactionHistoryBloc
 
               // Merge incoming batch by internalId, updating confirmations and other fields
               final Map<String, Transaction> byId = {
-                for (final t in state.transactions) t.internalId: t,
+                for (final t in state.transactions) _txKey(t): t,
               };
 
               for (final tx in newTransactions) {
                 final sanitized = tx.sanitize(myAddresses);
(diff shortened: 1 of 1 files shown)


mariocynicys:
is this deployed somewhere?

takenagain:
> is this deployed somewhere?

Deployments only run for PRs into specified branch prefixes, like `dev` in the base PR #3398, so there is unfortunately no preview link available for this PR.

smk762:
> This was resolved in KDF [#2691](https://github.com/KomodoPlatform/komodo-defi-framework/pull/2691), and SC/SCZEN transactions appear as expected for me in #3398 (once confirmed). I believe this PR can be closed, leaving `internal_id` as a unique identifier for transactions in the wallet and the SDK (luckily only one reference in each, so changing this behaviour is doable).

confirmed, closing

PR #3418, opened 2025-12-11
chore(macos): update Podfile.lock and project configuration

DeckerSU:
We should merge this, otherwise we will not be able to sign the macOS release with the existing Apple Developer account.


DeckerSU:
image
image

Screenshots of app and `.dmg` installer.

PR #3493, opened 2026-06-26
fix(macos): required build fixes for macOS (signing + DMG packaging)

DeckerSU:
## Summary

This PR contains the fixes required to build the project for macOS. Without
these changes the macOS build / packaging does not complete.

### Code signing — switch to Gleec-BTC OU team & GleecDEX bundle id
- Development Team `8HPBYKKKQP` (TillyHK Ltd.) → `B52ZCS7TMQ` (Gleec-BTC OU)
- Bundle identifier `com.komodo.wallet` → `com.GleecDEX.wallet`, aligning
  macOS with iOS
- Release-production now points at the
  `GleecDEX-B52ZCS7TMQ-Provisioning-Profile` Developer ID provisioning profile

Without the correct team / bundle id / provisioning profile the macOS target
fails to sign and the build cannot be produced.

### DMG packaging — fall back to system temp on `noowners` volumes
`hdiutil` cannot attach a DMG to a `-mountpoint` located on a volume mounted
with `noowners` (typical for external drives), failing with
`attach failed - Permission denied`. `contrib/make-dmg.sh` now detects when
`TMPROOT` resolves to such a volume and relocates the temp dir (and mountpoint)
to the system temp, which is always owners-enabled. Behaviour on the system
volume is unchanged.


macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -515,7 +515,7 @@
 					"$(PROJECT_DIR)",
 				);
 				MACOSX_DEPLOYMENT_TARGET = 13.5;
-				PRODUCT_BUNDLE_IDENTIFIER = com.komodo.wallet;
```
CharlVS, line 518: @DeckerSU  Will this mean that users' data from Komodo Wallet on macOS won't be migrated over to the new Gleec Dex app?

DeckerSU: > [@DeckerSU](https://github.com/DeckerSU) Will this mean that users' data from Komodo Wallet on macOS won't be migrated over to the new Gleec Dex app?

We have `com.GleecDEX.wallet` app in our AppStore account. Users of Komodo Wallet could migrate everything via export / import seed or file.

DeckerSU: The situation is exactly the same as with apps on other platforms. Neither for iOS nor for Android did we provide support for migrating from Komodo Wallet, as Gleec had its own applications. The situation here is exactly the same. Users who need to migrate data from Komodo Wallet -> Gleec DEX should use the standard export/import mechanism. We may need to write an article on how to export and import data.

CharlVS (commented):
@DeckerSU Please confirm that you've built and tested the signed macOS build locally, since I do not have the signing permissions to do so.

Beyond that, LGTM, and will be ready for approval after your confirmation.

DeckerSU:
> @DeckerSU Please confirm that you've built and tested the signed macOS build locally, since I do not have the signing permissions to do so.
> 
> Beyond that, LGTM, and will be ready for approval after your confirmation.

Yes, confirmed - everything was tested locally before opening the PR. The signed [.dmg](https://nebula.decker.im/gleecdex-macos-ccf47aea5.dmg) was built successfully and works as expected.

One thing to note: the build config should reference hash of the universal kdf binary rather than the static library (.a) for macOS - I mentioned this on internal resources, so you might not have seen it. Make sure that's reflected on your end as well.

CharlVS:
> One thing to note: the build config should reference hash of the universal kdf binary rather than the static library (.a) for macOS - I mentioned this on internal resources, so you might not have seen it. Make sure that's reflected on your end as well.

Thanks. Will do an update for this and then prep the release.

Issue #360, opened 2020-07-19
[FEATURE REQUEST]: Addressbook alphabetical sort


tonymorony:
It would be great to sort addressbook alphabetically without the app restart

naezith:
This is a back-end task.

tonymorony:
> This is a back-end task.

Sounds like dynamical redraw on new address adding task, afaik @Milerius already sort it on backend because on app restart it's rendering sorted

naezith:
> > This is a back-end task.
> 
> Sounds like dynamical redraw on new address adding task, afaik @Milerius already sort it on backend because on app restart it's rendering sorted

Data model / array is handled fully in back-end now. The data is directly fed to the `ListView`.

tonymorony:
became duplicate of https://github.com/KomodoPlatform/atomicDEX-Desktop/issues/534 I wonder.

Issue #358, opened 2020-07-13
[FEATURE REQUEST]: Pagination for DEX->History and for Wallet -> Transactions history


tonymorony:
Noticed that at the moment only last 50 trades / transactions displaying even if user got more.

It would be great to add pagination for these pages.

For swaps history call https://docs.komodoplatform.com/basic-docs/atomicdex/atomicdex-api.html#my-recent-swaps there is  `from_uuid` slicing param

For tx history call https://docs.komodoplatform.com/basic-docs/atomicdex/atomicdex-api.html#my-tx-history there is from_id slicing params

tonymorony:
probably this one is better to combine with "Orders - History" rework (filters + probable movement onto the orderbook widget to make it multi-page) so better to evaluate this one in next iter, unlikely we'll have time to include it into 0.2.1

tonymorony:
closing this one in favor of fliters

Issue #20, opened 2020-02-27
Local time


naezith:
All dates are currently UTC. We need to have them for local time, mine is UTC+3 for example. mm2 does not provide that, we need to add the timezone in backend on top of that timestamp I suppose.

tonymorony:
became dupl of https://github.com/KomodoPlatform/atomicDEX-Pro/issues/225

PR #3500, opened 2026-07-04
perf(wallet): cut HD login from minutes to seconds, plus TRON gas-free flows

CharlVS:
Two bodies of work on one branch: the **TRON gas-free wallet flows**, and a **wallet-load latency effort** that came out of users reporting balances taking *minutes* to appear after login.

---

### The problem, measured

Not one wall-clock number existed before this. So the first deliverable was a measurement that could not be argued with: `tool/kdf_latency_probe.py` spawns the KDF binary and speaks its JSON-RPC over HTTP using **Python standard library only** — no Flutter, no Dart, no SDK. A measurement taken *through* the SDK cannot tell you whether the SDK is the problem.

It found that a fresh HD login spends its time inside KDF, walking an HD address gap over the network, one address at a time:

| | measured |
|---|---:|
| KMD activate, `gap_limit: 20` (shipped default) | 46.9s |
| BTC-segwit activate | 121.2s |
| `enable_eth_with_tokens` (ETH + 2 tokens), HD | 363.8s |
| the app's own default coin set, HD | **480.7s (8 min)** |
| the same set, iguana | 8.9s |

Across the whole matrix, **0.3% of wall clock was spent inside HTTP calls** — the rest was the probe sleeping between polls while KDF worked on a task it had already accepted.

### The fix

`scan_for_new_addresses_impl` awaited one `is_address_used` per address, so `gap_limit: 20` cost **2 × 21 = 42 strictly serialized round trips per coin**. It now probes in **windows** sized to exactly the number of consecutive unused addresses still needed to end the walk — never probing an address the sequential walk wouldn't have, just not waiting for each answer before asking the next.

Results are consumed in order, driving the identical state machine. `test_scan_for_new_addresses` — which asserts the exact *ordered* list of probed addresses and the resulting counts across two chains and two accounts — **passes unchanged**. That is the equivalence proof.

KDF: `perf/hd-scan-concurrency` @ `ed8de236b`, built for all 7 targets and deployed.

### Measured through the real SDK

| | before | after |
|---|---:|---:|
| HD `activation_ms` | 35,966ms | **7,161ms** |
| HD time-to-first-balance | 47,986ms | **8,303ms** |
| `enable_utxo::status` polls | 93–95 | **15** |
| `account_balance::status` polls | 209–210 | **11** |
| iguana *(control)* | 2,566 / 3,149ms | 2,062 / 2,322ms |

HD is now **3.6× iguana instead of 25×**. The gap-limit cost also stopped scaling: 9.1 / 46.9 / 107.0s at gap 1/20/50 became **7.1 / 6.1 / 7.6s** — so `gap_limit` is now a correctness choice rather than a latency one.

`task::account_balance` fell 15.9× **with no change to its code**: it had been blocking on the whole-wallet `HDAccountsMutex` held by the still-running scan.

### Correctness fixes found along the way

- **`hydratedPubkeys` could return the wrong wallet's pubkeys.** It read `_pubkeysCache` *before* `_captureWalletContext()` — the call whose side effect detects an undelivered wallet switch and clears that cache. `BalanceManager` paints a balance straight from the result. Regression test added, and verified to fail without the fix.
- **A silent progress stream hung the caller forever.** `return completer.future` sat after the `await for`, so a stream that never emits *and* never closes suspended before the caller ever received the future; the deadline fired and the caller still waited. Joiners were unaffected, which hid it.
- **The activation deadline was below measured reality.** A flat 60s fired mid-activation on every fresh HD login, and the retry issued a duplicate concurrent enable. Now protocol-aware (8 min EVM / 3 min otherwise, ZHTLC exempt), with the ordering invariant against `CoinsRepo._activationAttemptTimeout` documented in both files.
- **KDF panicked on TRX/NFT activation with P2P disabled** — an `unwrap()` on a P2P context fetched *before* checking whether proxy signing was needed. Fatal on wasm32, where `catch_unwind` is unusable with async. Fixed at all three call sites.
- **`GetPublicKeyHashRequest` could never be sent.** `'params': <JsonMap>{}` — a single type argument on `{}` makes an empty **Set**, which `jsonEncode` refuses. Present since 2024, invisible to the type system and to mock backends.
- **The login form auto-filled the password from the clipboard.** Removed.

### Regression protection

- `komodo_defi_harness` — a real `KomodoDefiSdk` with only the RPC backend faked, so auth/activation/pubkeys/balances/storage are production code. Replay tier gates every PR; a real-KDF process tier runs nightly.
- 18 test files existed but were never registered in `test_units/main.dart` and had never run. Registering them surfaced real failures, fixed here.

---

## 2. TRON gas-free wallet flows

Gas-free TRC-20 send and receive via the custody-address model, the GasFree relay integration, and the associated UX.

---

## Test status

| suite | result |
|---|---|
| app `test_units` | 528 pass / 3 skip |
| `komodo_defi_sdk` | 631 pass / 1 skip |
| harness (replay) | 21 pass |
| harness (process, real KDF) | 4 pass |

## Reviewer notes

- **Generate lockfiles with the pinned CI Flutter (3.41.4), not whatever is on PATH.** A lock generated on 3.44.0 is unsatisfiable under `--enforce-lockfile` and fails every job before it does any work. Nothing in the repo pins this locally — worth adding an `.fvmrc`.
- The macOS Xcode project is rewritten by any Flutter build that has SwiftPM enabled (`flutter config --no-enable-swift-package-manager` fixes it *per machine only*), and such a build also truncates `macos/Podfile.lock` from 61 pods to 5. Neither should ever be committed.
- Full measurement method and raw numbers: `docs/KDF_LATENCY_REPORT.md`. KDF-side analysis and what remains: `docs/KDF_IMPROVEMENT_OPPORTUNITIES.md`.
- **Biggest remaining win is KDF-side and not in this PR:** `try_rpc_send` holds the `web3_instances` mutex across the network round trip, capping EVM at one RPC in flight per coin. EVM activation is still 196.9–346.4s.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


CharlVS:
## Update: KDF rolled to `main` `f3efd2c`, descoped from the three open perf PRs

Pushed `524f0654`…`a6caa91e`. This changes what the branch ships, so the title and §1 of the description above are now **ahead of the contents**.

**The pin moved** from `538724e9e` (`perf/evm-rpc-429-backoff` — the top of kdf-internal [#18](https://github.com/GLEECBTC/kdf-internal/pull/18)/[#19](https://github.com/GLEECBTC/kdf-internal/pull/19)/[#20](https://github.com/GLEECBTC/kdf-internal/pull/20), none merged) to `f3efd2ca10420f2982fa127dde84dcc17891f577` (`main`, `3.1.0-beta_f3efd2c`), which carries the merged gas-free work.

So **the HD-login speedup described above is not in this branch right now.** `main` has no concurrent gap scan and no EVM RPC pooling — expect login to feel like the pre-perf builds. Verified in the shipped binary: zero `web3_pool` symbols. If #18/#19/#20 land before this merges and the pin is rolled forward, the original claim becomes true again; until then it does not, and testers should be told so.

**The gas-free flows are unaffected.** The gas-free module is byte-identical between the two pins apart from one file, with no serialized type changed — and more broadly the RPC wire contract is identical: zero serde/`rename`/`deny_unknown_fields` changes anywhere in the non-test diff, byte-identical dispatch table.

**Only one GUI improvement was actually reliant on the perf PRs: ws-on-web.** Web was sending `wss://` node entries (1100 nodes in the shipped config carry a `ws_url`) to a KDF that panics on a reply arriving after the caller's 10s timeout but inside the 30s notifier window — and wasm cannot `catch_unwind`. Now gated behind `_kSendWsNodesOnWeb`. Electrum's WSS path is unaffected, so UTXO keeps websockets on web.

**Three things the roll broke that are *not* descopes** and must survive the perf PRs landing: `main`'s Windows artefact ships three proc-macro DLLs that fail the extraction guard (broke every build and `flutter test`); `api_build_config_test.dart` pinned a literal commit and had been red for three rolls; and `source_urls` had to lead with the only mirror serving `main/`, because the preview workflow takes `source_urls[0]` with no fallback under `--strict`.

Also fixed here: the macOS Runner **Release** config alone carried `com23.GleecDEX.wallet` and team `34Y7669588`, a stray partial revert of #3493.

### Verification

| suite | result |
|---|---|
| `komodo_wallet_build_transformer` | 113 passed |
| `komodo_defi_rpc_methods` | 236 passed |
| `komodo_defi_sdk` | 851 passed, 1 skipped |
| `test_units/main.dart` | 603 passed, 3 skipped |

Built and confirmed to carry `3.1.0-beta_f3efd2c`: web `--release`, Android debug APK, iOS device `--no-codesign`, macOS unsigned. Signed macOS untested (no local provisioning profile), Windows/Linux untested (no host). The iOS **simulator** cannot build — both pins ship an identical device-only `libkdf.a`, so that is pre-existing.

Full rationale, accepted regressions, and the re-introduction checklist: [`docs/KDF_PERF_STACK_DESCOPE.md`](docs/KDF_PERF_STACK_DESCOPE.md). SDK side: GLEECBTC/komodo-defi-sdk-flutter#360.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

CharlVS:
> **Correction to my comment above.** I wrote that "the HD-login speedup described above is not in this branch right now." That was too broad and I got it wrong. The **KDF-side** speedup is gone; the **client-side** speedup is not, and it is substantial. Details below; the rest of that comment stands.

## What survives the roll, and what does not

The latency work in §1 of the description has two halves. Only one of them was in KDF.

**Still shipping — eight client-side commits, none of which depend on the KDF pin:**

| commit | what it does |
|---|---|
| `28122573` | bound the repeating HD gap scan and **scale its limit** |
| `8c1fa31f` | skip the redundant post-activation HD address scan |
| `3fc991a8` | cut identity RPCs, paint balances from storage |
| `935552af` | yield cached transactions before waiting on activation |
| `99312d68` | stop paying for work the RPC and storage paths throw away |
| `1fbbc8ba` | share one EVM balance streamer |
| `c5aab12fe` | cut redundant work out of the login activation storm |
| `aa78d4510` | render the tx list from cache instead of waiting on pubkeys |

**The gap-limit reduction is the big one, and it attacks exactly what `main` is slow at.** `HdGapLimit.resolve` (wired at `activation_manager.dart:562` and `pubkey_manager.dart:395`) sends **3** for software wallets and **1** on a newly generated wallet's first sign-in, against KDF's default of 20.

`docs/KDF_LATENCY_REPORT.md` §4a measured that on `bd413dc` — the *sequential* KDF, which is what `main` is for this path — single-coin KMD HD activation:

| `gap_limit` | activate |
|---|---:|
| 20 (KDF default) | 46.9s |
| 3 (what we now send) | ~13.1s *(modelled at the measured ~2.0s/gap unit)* |
| 1 (generated, first sign-in) | 9.1s |
| `do_not_scan` (floor) | 2.0s |

So roughly **3.6x off a single-coin HD activation, from the client, with no KDF change** — 46.9s → ~13s. The perf stack reached 6.1s at gap 20; the client-side change gets most of that distance on `main` alone.

**Two exceptions worth knowing:**
- **Trezor does not benefit.** `HdGapLimit.resolve` returns `hardware = 20` for `PrivateKeyPolicy.trezor()`, so hardware wallets still walk the full gap sequentially.
- **TRX does not benefit.** `TrxWithTokensActivationParams` has no `gapLimit` field at all, so TRON activation always walks KDF's default 20. (Recorded as §8.1 in the descope doc.)

**What is genuinely lost with the roll** — all KDF-internal, none recoverable client-side: the *concurrent* gap scan (probes in parallel rather than one at a time), EVM RPC parallelism and HTTP connection pooling, and the 429 backoff/retry for EVM and TRON.

**Net:** login is slower than the perf-stack builds testers last saw, but materially faster than the pre-optimisation baseline this work started from — and first paint in particular is better regardless of KDF, because balances and transaction history now render from storage instead of waiting on activation.

That also means the PR title is closer to true than my earlier comment implied. I have still left it alone: if kdf-internal #18/#19/#20 land and the pin rolls forward, it becomes fully accurate again.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

CharlVS:
## Ready for review — stack reviewed, fixed, and green

Marked ready. The SDK side lands in three ordered PRs; this PR's submodule points at the top of that chain.

```
sdk/dev
  └─ add/tron-gas-free ..................... GLEECBTC/komodo-defi-sdk-flutter#358
       └─ perf/hd-gap-scan-and-ws-transport ..... #361
            └─ chore/roll-kdf-main-descope ...... #360
```

**Merge them bottom-up (#358 → #361 → #360), then this.** #358 must not land alone: its degraded-identity resets are only cured by the guard in #361.

### Fixes made since the last update

A multi-agent review of all 77 SDK commits produced 44 findings, each adversarially verified. Seven landed as fixes:

| | fix |
|---|---|
| **BLOCKER** | `getActiveUser()` cleared auth on *any* throw, so one dropped loopback socket called `kdfStop()` and forced a sign-out — from a read accessor that `isSignedIn()` and several managers poll. Guarded on `AuthException(apiConnectionError)`; matching the raw transport types would never have fired, because `_runStartupSensitiveRpc` retries and rewraps them. |
| **BLOCKER** | Transaction history forked its storage namespace on a transient `get_public_key_hash` failure — the enriched-prefix rows were then orphaned and swept. Silent loss of persisted history. It was the one wallet-scoped manager missing the `isDegradedWalletIdentity` branch. |
| **MAJOR** | The withdraw path passed a null pubkey hash into `ensureWalletSession`, wiping the gas-free capability session mid-money-flow. |
| **MAJOR** | An unbound identity latched the terminal `securityMismatch`, permanently disabling gas-free for an asset whose only problem was not being bound yet. |
| **MAJOR** | A trace-less journal reservation had no way out, permanently blocking gas-free sends for that custody address. Added an explicit `discardPendingGaslessTransfer`. An expiry sweep was tried and **rejected** — it would have allowed a possibly-relayed transfer to be sent twice, which `legacy record without trace stays unknown and non-resubmittable` correctly catches. |
| **MAJOR** | Trezor lost the device's real error message to a deliberate `toString()` redaction. Fixed at the throw site; the test was right and is unchanged. |
| **MAJOR** | The SSE preflight test could never pass — its fake server never flushed headers, so the client correctly waited forever. |

Two regression tests were each verified to **fail without their fix**.

### Verification

All suites green: `komodo_defi_sdk` 852, `komodo_defi_local_auth` 59, `komodo_defi_framework` 39, `komodo_defi_rpc_methods` 236, `komodo_wallet_build_transformer` 113, app `test_units` 603.

CI on this PR is green on **16 of 17** checks, including Windows/macOS/Linux desktop, Android, iOS, web-app-macos, both replay tiers and the preview deploy.

### The one red check is not ours

`Test web-app-linux-profile` fails, and has failed **repo-wide since 2026-08-12** — including on `dev` itself (`a9ac5e227`) and on every other branch (`feat/nav-nft-and-onboarding`, `improve-release-sdk-tag-check`). The driver log shows the web app never starts KDF (`[KomodoDefiSdk] KDF is not running` every 5 minutes for 15 minutes) until the 20-minute `request_data` timeout. `dev` is not branch-protected, so it gates nothing mechanically — but it deserves its own issue rather than being carried into the RC unexamined.

### Known trade-off for the RC

HD login will be **slower than the perf-stack builds** and faster than the pre-optimisation baseline: the client-side gap-limit reduction survives the roll, the KDF-side concurrency does not. See `docs/KDF_PERF_STACK_DESCOPE.md` §4.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #16, opened 2020-02-06
[Orders]: Backend/Frontend


Milerius:
- In the backend we have:

```cpp
        Q_INVOKABLE QVariantMap get_my_orders();
```

Which is a json / javascript object that look like:

```json
{
  "BTC": {
    "orders": {
      "maker_order": [
        {
          "rel": "KMD",
          "base": "BTC",
          "date": "19 Janvier 2020",
          "cancellable": true,
          "base_amount": "1",
          "rel_amount": "1",
          "order_id": "4321"
        }
      ],
      "taker_order": [
        {
          "rel": "ETH",
          "base": "BTC",
          "date": "20 Janvier 2020",
          "cancellable": true,
          "base_amount": "1",
          "rel_amount": "1",
          "order_id": "1234"
        }
      ]
    }
  }
}
```

The C++ struct is:

```cpp
struct qt_my_order_contents : QObject
    {
        Q_OBJECT
      public:
        explicit qt_my_order_contents(QObject* parent = nullptr);
        QString m_order_id;
        QString m_date;
        QString m_base;
        QString m_rel;
        bool    m_cancellable;
        QString m_base_amount;
        QString m_rel_amount;
        QString m_price;

        Q_PROPERTY(QString price READ get_price CONSTANT MEMBER m_price)
        Q_PROPERTY(QString date READ get_date CONSTANT MEMBER m_date)
        Q_PROPERTY(QString base READ get_base CONSTANT MEMBER m_base)
        Q_PROPERTY(QString rel READ get_rel CONSTANT MEMBER m_rel)
        Q_PROPERTY(bool cancellable READ is_cancellable CONSTANT MEMBER m_cancellable)
        Q_PROPERTY(QString base_amount READ get_base_amount CONSTANT MEMBER m_base_amount)
        Q_PROPERTY(QString rel_amount READ get_rel_amount CONSTANT MEMBER m_rel_amount)
        Q_PROPERTY(QString uuid READ get_uuid CONSTANT MEMBER m_order_id)

        [[nodiscard]] QString get_uuid() const noexcept
        {
            return m_order_id;
        }

        [[nodiscard]] bool
        is_cancellable() const noexcept
        {
            return m_cancellable;
        }

        [[nodiscard]] QString
        get_base_amount() const noexcept
        {
            return m_base_amount;
        }

        [[nodiscard]] QString
        get_rel_amount() const noexcept
        {
            return m_rel_amount;
        }

[...]

Issue #10, opened 2020-01-05
[UI/UX] prototype layout


ca333:
For UI layout and UX please orientate towards these mockups: 

![dextop_wallet_view](https://user-images.githubusercontent.com/10762374/71768757-a492e680-2f19-11ea-8f13-19e93320b025.jpeg)

![dextop_DEX_dropdown01](https://user-images.githubusercontent.com/10762374/71781691-b041e400-2fd1-11ea-8c45-c6d214915d09.jpeg)

![dextop_DEX_dropdown02](https://user-images.githubusercontent.com/10762374/71768759-abb9f480-2f19-11ea-8dd8-617c0f5c4037.jpeg)

![dextop_DEX](https://user-images.githubusercontent.com/10762374/71768763-b07ea880-2f19-11ea-9df0-3e046befbe64.jpeg)

ca333:
Font: https://fonts.google.com/specimen/Rubik?selection.family=Rubik

ca333:
Alternative `rel` drop-down: 

![dextop_DEX_dropdown02_alternative](https://user-images.githubusercontent.com/10762374/71784273-1f2f3500-2ff2-11ea-9640-3dcdf715c3b6.jpeg)

naezith:
Current version covers all these now.

PR #3504, opened 2026-07-21
fix(swap): harden trading lifecycle and execution

CharlVS:
## Preview purpose

Deploy the adversarially audited Swap surface for product and runtime validation. This is a preview-only, stacked branch based on the work already present in #3501; it is not a release approval.

## What changed

- bind quotes, previews, reviews, activations, orders, bot lifecycle operations, cancellations, and recovery to the exact live wallet/session and input revision
- add immutable route consent, fee/source/recipient checks, parser limits, exact-once mutation guards, and fail-closed uncertain-outcome handling
- harden maker/taker previews, Market Maker ownership/configuration, strict settings persistence, and cross-wallet cache behavior
- apply the agreed UX behavior: reversing clears the amount and legacy-link amounts remain drafts until **Check routes**
- pin this preview to the companion SDK preview commit and development KDF artifact manifest

## Validation

- changed Dart files: zero analyzer findings
- whole repository: zero analyzer errors; remaining findings are confined to unchanged legacy files
- nested SDK route manager analysis: no issues
- Dart formatting, translation JSON, and repository diff checks: clean
- tests were not run, per the repository instruction that the existing unit/integration suites are currently failing

## Preview and release safety

The Firebase preview workflow enables Unified Swap against the configured preview proxy. The KDF development artifact pin is intentionally preview-only. Production remains blocked on immutable production artifacts and the outstanding release-readiness gates.


CharlVS:
Closing as superseded by #3507, which implements the same feature against the contract KDF is actually building.

## Why this cannot be revived as-is

The LI.FI work here targets `experimental::trade_route::*`. That namespace's KDF side was closed unmerged across kdf-internal #10–#16 on 2026-07-18 and the branches were deleted. It was replaced by `routed_swap::` ([gleec-specs#2](https://github.com/GLEECBTC/gleec-specs/pull/2)). `grep routed_swap` over this branch returns 0 hits, against 7,439 lines of models keyed to the dead names.

Two other things here should not merge regardless of the swap decision:
- `sdk/…/app_build/build_config.json` pins KDF to branch `local/lifi-latest` from `kdf-dev-builds.nitride.app`, having removed the three trusted sources (GitHub releases, `devbuilds.gleec.com`, `nebula.decker.im`). The LI.FI proxy is hardcoded to the same personal domain.
- `pubspec.yaml` ships `prototypes/gleec-unified-swap/…-assets/` into the production app bundle.

## What was triaged out before closing — worth keeping

Commit `a2985ce32` bundles ~10,583 lines of legacy DEX/trading changes across 77 files in with the swap work. Most is churn (`super.key`, formatting), but it contains real hardening that is **not** LI.FI-coupled and is **absent from `dev`**:

- **Exactly-once guard on swap recovery** — `canRecoverSwap` (10 call sites) plus a `RecoverySubmissionStatus` enum with an explicit `uncertain` state, streamed from `TradingEntitiesBloc`. Today the recover button can be submitted twice.
- **Exactly-once guard on order cancellation** — `canCancelOrder` (9 call sites), checked in the button and again in the bloc before the RPC.
- **Wallet/session binding** — `kdf_wallet_authority.dart`, `market_maker_bot_wallet_session.dart`, `trading_entity_id.dart`, so trading operations are bound to the exact live wallet rather than whatever is current when a response lands.
- **Order model validation** — `order_model_validation.dart`.

None of that was ported into #3507, which deliberately left the legacy DEX untouched. It should be extracted into its own PR against `dev`, reviewed on its own terms — it is a trading-correctness change, not a swap feature, and it was never reviewed or tested here (this PR's own description notes tests were not run).

Reopen if that extraction turns up anything else worth recovering.

Note: #3501 (Gnosis Pay card), which this was stacked on, is unaffected and stays open.

PR #3507, opened 2026-08-12
feat(swap): LI.FI unified swap (preview)

CharlVS:
## What this is

A preview of the LI.FI unified swap feature, built against the `routed_swap` contract proposed in [GLEECBTC/gleec-specs#2](https://github.com/GLEECBTC/gleec-specs/pull/2).

**Draft, and it cannot fully work yet.** KDF does not implement `routed_swap` — `shamardy` has the LI.FI HTTP client and the contract is still in review. Everything here runs against a scripted fake KDF. What is previewable today is the atomic (peer-to-peer) path and the whole UI surface; routed quotes will return errors until KDF ships.

Based on `add/gas-free-tron` so the diff is only this work.

## Why not the earlier implementation

The previous LI.FI program targets a different KDF contract from the one in review:

- It is written against `trade_route::` RPCs, which KDF does not implement.
- The contract in review is a smaller, different one (`routed_swap::`), already approved by `mariocynicys`.
- The Flutter side is welded to that namespace: `grep routed_swap` over that branch returns **0 hits**, against 7,439 lines of hand-written models keyed to `experimental::trade_route::*`. The prototype names it as its authority too.

See #3504 for that work. Nothing was ported; the policy layer, the ~129 tests and the prototype's verifier were used as a statement of intent and re-derived.

## Structure

| Layer | What it does |
|---|---|
| `komodo_defi_harness` | Scripted fake KDF for the whole `routed_swap` contract |
| `komodo_defi_rpc_methods` | `client.rpc.routedSwap` — typed bindings |
| `komodo_defi_sdk` | `sdk.routedSwaps` — hides the task lifecycle entirely |
| `lib/shared/swap/` | Source-agnostic quoting and execution over routed + atomic |
| `lib/bloc/unified_swap/` | Form, re-price consent, progress |
| `lib/views/swap/` | Swap / Activity / Advanced |

## Navigation change

The Swap menu entry now opens the unified form, with the full trading interface moved to an **Advanced** destination in the same shell. Nothing is removed and a test asserts Advanced stays reachable.

This is the same arrangement #3504 shipped, but that one had the swap feature behind dart-defines defaulting to `false`, so the default destination rendered "Unified Swap is unavailable" and the only working trading UI was the one behind the sub-tab. There is no such gate here.

`/dex/trading_details/<uuid>` opens directly on Advanced, since only the trading interface renders it. No route parser changes were needed.

## Product constraint worth knowing

LI.FI supports 70 EVM chains and **the Gleec chain (11169) is not one of them**. Cross-referencing the coin config against its token list: **340 of our 727 assets are routable**; 196 are non-EVM, 144 are on supported chains but unlisted, and 47 are on unsupported chains — including GLEEC, GLEECT and the 8 GRC-20 tokens.

Routed swaps can never move our own native asset. That is why the atomic orderbook stays a first-class source on the same screen rather than something routed swaps replace.

## Decisions in the code worth reviewing

- **Ranking compares guaranteed amounts, not expected ones.** A routed quote's headline is subject to slippage while an atomic fill is not, so ranking on headlines would systematically promote the looser promise.
- **Pressing start re-prices first**, and any drop in the guaranteed amount stops for consent with old and new side by side. The threshold is zero, not a tolerance band.
- **Screens never over-claim.** The headline is the guaranteed receive; an incomplete fee total says so (the quote omits approval gas); a refund and a partial fill are terminal but never rendered as completed; "your balance is unchanged" appears only where the contract proves it.
- **Atomic orders are fill-or-kill.** A resting or partially filled order is a different trade from the one that was quoted.

## Testing

- 617 app tests, 80 SDK tests, zero analyzer errors.
- No live-KDF coverage: the fake is scripted from the spec, so it is two readings of an unimplemented contract. A round-trip test parses the fake's output with the real SDK models so the two cannot drift apart unnoticed.

## Known gaps

- `routed_swap::history` has **no published schema** (spec §7). Its shape here is our review proposal, isolated in one file and marked provisional. The history screen should not ship until §7 lands.
- Activity shows routed swaps only; the atomic list still lives under Advanced.
- `SwapInfo.isSuccessful` is used as the SDK defines it (`isComplete && errorEvents.isEmpty`). Worth confirming — in KDF, `error_events` reads more like the list of *possible* error events than ones that occurred. If so the getter is wrong and every consumer is affected, not just this one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


CharlVS:
### UI tests: the remaining failure is an ops credential issue, not code

`ui-tests-on-pr` is red on both runners, but the swap-shell navigation fix in 594b12249 did its job. Suite progression is identical on `web-app-linux-profile` (37m) and `web-app-macos` (41m):

| Suite | Result |
| --- | --- |
| `wallets_tests` | pass |
| `wallets_manager_tests` | pass |
| `dex_tests` | **pass** |
| `misc_tests` | **pass** |
| `fiat_onramp_tests` | **fail** |

`dex_tests` completing means the maker and taker flows navigated the new Swap surface and ran a real atomic swap against the counterparty. Per this workflow's own comment, `misc_tests` and `fiat_onramp_tests` had not executed on any branch since 2026-08-12 — they sit behind `dex_tests` in a fail-fast loop. So `fiat_onramp_tests` is being **reached** for the first time, not newly broken.

#### The failure

```
Expected: exactly one matching candidate
  Actual: _KeyWidgetFinder:<Found 0 widgets with key [<'fiat-payment-method-ramp-0'>]: []>
```

Probing `fiat-ramps.gleec.com` directly with the test's exact selection (EUR → LTC, 50):

| Call | Result |
| --- | --- |
| Ramp `GET ?endpoint=/currencies` | 200, real data |
| Ramp `GET ?endpoint=/assets` | 200, real data |
| Ramp `POST ?endpoint=/onramp/quote/all` | **`AUTH.ACCESS.INVALID_HOST_API_KEY`, `statusCode: 403`** |
| Banxa `GET ?endpoint=/api/payment-methods&source=EUR&target=LTC` | 200, a real Visa/Mastercard method |

The Ramp host key the proxy sends is rejected at the quote endpoint. Banxa is healthy — the test would have passed its second assertion.

Two things make this look like a UI bug when it isn't:

1. **The proxy answers HTTP 200 and carries Ramp's 403 in the body.** A status-code health check on that host looks green.
2. **Only the quote endpoint needs the host key.** `/currencies` and `/assets` answer fine, so the fiat form loads and both dropdowns populate; the failure appears only at the payment-method step.

The `credentialHint` already in `test_integration/tests/fiat_onramp_tests/form_tests.dart` names this correctly.

#### What this needs

A Ramp host-key rotation on `fiat-ramps.gleec.com` — outside this repository. Until that happens this suite cannot pass, so the UI gate will stay red on this PR regardless of its contents. Deliberately not skipping the Ramp assertions: that would hide a real outage in a user-facing flow.

PR #3514, opened 2026-08-27
fix(fiat): validate on-ramp checkout URLs before framing them

CharlVS:
## Summary

`assets/web_pages/fiat_widget.html` base64-decoded its `fiatUrl` query parameter straight into the on-ramp iframe's `src` with no validation of any kind, so any URL reaching that parameter was loaded inside the wallet's own origin — including non-`http(s)` schemes, which execute as wallet-origin script with access to `localStorage` and IndexedDB (encrypted wallets, wallet list, KDF `rpc_password`).

The parameter is reachable two ways:

1. A crafted link to the wrapper page.
2. The server-supplied `checkout_url` from the fiat-ramps order response, which `FiatFormBloc` wrapped unconditionally with only an `isEmpty` check.

The second path matters most: the wrapper page is fetched from production by desktop and mobile builds that are **already in users' hands** — their `getOriginUrl()` is hardcoded to `https://dex.gleec.com`. That is also why the check has to live in the deployed asset and not only in the Dart caller: redeploying the asset protects shipped native clients with no app release.

## Changes

- **`lib/bloc/fiat/fiat_checkout_url_allowlist.dart` (new)** — `isAllowedFiatCheckoutUrl` requires `https`, rejects embedded credentials, rejects characters that Dart's URI parser and the browser's WHATWG parser read differently, and requires a host belonging to a payment provider.
- **`fiat_widget.html`** — carries the same lists and enforces them itself. Parses with `new URL()` (no base argument, so a relative URL fails rather than resolving against the wallet origin), and assigns the re-serialised `URL.href` so the validated string is the loaded string. A rejected link renders a visible explanation instead of a blank page.
- **`FiatFormBloc`** — validates the order response's `checkout_url`, and only routes through the wrapper page for providers that report status via `postMessage`. Banxa reports status by polling, so it now opens directly and never touches the wrapper. The provider is captured before the order request goes out, so a payment-method change while the request is in flight cannot route the response through the wrong transport.
- **wire encoding** — the payload is percent-encoded now, because a `+` in the query string is decoded as a space by the wrapper's `URLSearchParams` and silently corrupts it. It stays *standard* base64 rather than base64url: the wrapper page a build actually talks to is the one deployed at `getOriginUrl()`, and that page calls `atob` directly, which rejects base64url's `-` and `_`. The new decoder accepts standard base64, base64url, and the space-corrupted old form, so every combination of shipped client and deployed page decodes.
- **iframe sandbox** — dropped `allow-top-navigation`, and corrected `allow-top-navigation-by-user-action` to `allow-top-navigation-by-user-activation`. The former is **not a sandbox flag**; browsers discard the token, so the file's only working top-nav grant was the unconditional one. Provider "return to wallet" buttons keep working; a script inside the frame can no longer navigate the window on its own.
- **`postMessage`** — pinned to the wallet origin instead of `"*"` in `fiat_widget.html`, `checkout_status_redirect.html` and `bitrefill_widget.js`, and inbound status events are only relayed when they come from a provider origin.
- **`firebase.json`** — `frame-ancestors 'self'`, `X-Frame-Options`, `X-Content-Type-Options`, `Referrer-Policy`, a narrow `Permissions-Policy`, and `must-revalidate` on `/assets/assets/web_pages/**` so a future fix reaches shipped native clients immediately rather than after Firebase's one-hour default.

## Two deliberate non-changes

**`camera`, `microphone` and `payment` are left out of `Permissions-Policy`.** Their default allowlist is `self`, and that is exactly what permits the wrapper page to delegate them to the provider iframe via its `allow` attribute. `camera=()` would break provider KYC.

**No `connect-src` / `script-src` CSP.** `frame-ancestors` closes the framing gap and cannot white-screen the app. A full CSP needs a staging soak against CanvasKit, the KDF WASM module (`'wasm-unsafe-eval'`) and several hundred RPC hosts — worth doing, but not in a fix that should ship quickly.

## Allowlist scope

Provider hosts are matched by domain suffix (`*.banxa.com`, `*.ramp.network`) rather than exact host, because the partner sub-domain is inherited config and is not confirmed against a live `checkout_url`. An attacker still has to control provider infrastructure, where they would already control the checkout page itself. Tightening to exact hosts is a one-line change in the allowlist once someone confirms the real host.

## Testing

- 18 tests in `test_units/tests/fiat/fiat_checkout_url_allowlist_test.dart`. Beyond the allowlist itself they pin the things that can silently stop being true: that the Dart and HTML copies of the allowlist have not **drifted apart**, that the shipped asset still contains each guard, that the page deployed to users can still decode what this code sends it, and that `firebase.json` still sets the header *values* — not just the keys — that the verifier checks for.
- `flutter test test_units/main.dart`: 257 passed, 3 skipped, 0 failed. `flutter analyze` clean on every touched file.
- Verified in a browser against the built asset served on its own origin. Control first: the pre-fix file loaded the injected URL and its script's write landed in the wallet origin's `localStorage`. Same input against the fixed file: nothing executes, no `src` is set, the blocked panel renders. All hostile variants tried (wrong scheme, attacker origin, look-alike host, suffix trick, embedded credentials, parser-differential host) are rejected; all legitimate provider URLs still load, and both the old and new encodings decode correctly.

- Re-verified against this PR's own preview deploy, which is the path a real client takes rather than a local file: a `javascript:` payload sets no `src` and writes nothing to `localStorage`, `app.ramp.network.attacker.example` is blocked, an old-client standard-base64 payload containing `+` decodes correctly, and the browser accepts all five sandbox tokens. `tool/verify_web_deploy.sh` passes against the preview and fails every check against `dex.gleec.com`, which still serves the pre-fix page byte-for-byte.
- Firebase applies the headers of *every* matching `headers` entry, so the wrapper's `must-revalidate` genuinely overrides the `**` block rather than being shadowed by it. Measured on the preview, not assumed.

Browser testing is also what caught the invalid sandbox flag — it only shows up as a console error when the page is actually loaded.

## Hosting: where this actually has to land (second commit)

The headers only attach to the hosting config that is deployed, and nothing in this repo said which site serves production. Measured from the live sites:

| URL | Site | **Project** | Version | Last deployed | Deployed by |
|---|---|---|---|---|---|
| `dex.gleec.com` | `gleec-wallet-official` | **`gleec-wallet-official`** | 0.9.4 | 2026-04-17 | by hand |
| `walletrc.web.app` | `walletrc` | `komodo-wallet-official` | 0.9.6 | 2026-08-12 | CI, on `dev` |
| `prodrc.web.app` | `prodrc` | `komodo-wallet-official` | 0.9.3 | 2026-02-11 | nothing |

**Production is a different Firebase project from everything CI touches**, so no CI job that deploys the RC can also deploy production. The `main` → `prodrc` step in `firebase-hosting-merge.yml` has never run: the workflow only triggers on pushes to `dev`.

This matters more than usual here, because desktop and mobile builds already in users' hands fetch the on-ramp wrapper from `https://dex.gleec.com` at runtime — **deploying production is what fixes them, with no app release**. Until then they keep fetching the old page regardless of what has merged.

Verified live: `dex.gleec.com` currently serves the wrapper page byte-identical to the pre-fix file on `dev`, with no security headers.

So the second commit:

- makes production reachable from this repo's hosting config at all (see **Merging `dev`** below for the mechanism, which changed);
- adds `tool/verify_web_deploy.sh`, which asks a deployed site whether it is hardened — checking the wrapper asset and the headers separately, since they ship through different mechanisms. It needs only `curl`, so whoever holds production access can confirm the result without granting access to anyone else;
- runs that verifier in CI after the RC deploy;
- records the topology, the deploy commands, and how to re-derive it all (`/__/firebase/init.json`) in `docs/WEB_HOSTING_TOPOLOGY.md`.

```bash
tool/verify_web_deploy.sh https://dex.gleec.com
```

## Review round (third commit)

A multi-agent review of the two commits above, every finding put to an adversarial verifier. What survived:

- **The wire encoding was wrong in the one direction this PR is about.** base64url is not decodable by the page deployed at `dex.gleec.com` today. Whether a given checkout URL trips it is not random: for ASCII input, base64 index 62/63 can only come from `>`, `?` or `~` sitting at an offset ≡2 (mod 3), so it turns on the provider host's length — inherited config that can change without notice. `komodo.banxa.com` and `app.ramp.network` happen to be safe; `a.banxa.com` is not. The same analysis retires the original justification: a `+` needs a `>` or a `~`, so "1 in 64 characters" was wrong by a wide margin, and `Uri.encodeComponent` is what actually fixes that bug — on the deployed page as well as the new one.
- **`FiatFormBloc` read the selected provider back after the `await`.** Event handlers run concurrently and `_updatePaymentMethods` falls back to `methods.first` on a refresh, so the wrapper/no-wrapper decision could be made for a different provider than the one that produced the URL — putting a Ramp checkout straight into the webview, where its `postMessage` status relay does not exist.
- **Four guards reported green while the thing they guard was gone.** The wire-format test's fixture encoded byte-identically under both alphabets, so it could not see the encoding at all. The drift test accepted a double-quoted entry, a commented-out entry and a post-declaration `.push()` — all three fail it now. The `firebase.json` tests asserted header keys, so `frame-ancestors *` passed; so did the verifier. The wildcard-`postMessage` guard matched only the double-quoted spelling the file no longer uses.
- **`verify_web_deploy.sh` turned one network blip into five "header is missing" failures**, with guidance pointing at `firebase.json`. The header probes are retried and cache-busted now, and a transport failure is reported once, with curl's own message.
- Smaller: the wrapper's inbound message gate requires `https` and not merely an approved host; the inert `interest-cohort` token is gone from `Permissions-Policy`; the new CI step has an `id`.

Recorded so they are not re-raised — findings that did **not** survive verification: the framed provider navigating itself after load is origin-isolated and loses its `allow` delegation (an `allow` list with no explicit origin defaults to `'src'`); the blocked panel needs no focus management, because it is the document's entire content once the iframe is hidden; and the topology doc's `main.dart.js` claim is correct, since the deployed bundle does carry the real project values.

## Merging `dev` (fourth commit)

`dev` moved on while this was open, and PR #3500 changed the same file in an incompatible direction: it turned the single hosting entry from `"site": "walletrc"` into `"target": "walletrc"` and taught `.firebaserc` to map that target per project (`komodo-wallet-official` → site `walletrc`, `komodo-wallet-preview` → site `komodo-wallet-preview`).

That is the mechanism Firebase provides for "one config, a different site per project" — which is exactly the problem this PR had solved by hand, badly. Keeping the two-entry array would have been wrong twice: it reintroduces a hand-maintained duplicate of the one block that must not drift, and pinning `site: walletrc` would have broken the preview indirection #3500 had just added, so a local `firebase deploy` would resolve to the RC site name rather than the preview site.

So the array is gone. There is **one** hosting entry, and `.firebaserc` gains a third block mapping `walletrc` → site `gleec-wallet-official` under project `gleec-wallet-official`. Production and the RC now deploy the same config by construction; the headers cannot drift, and the test that policed the duplication is deleted rather than maintained.

Resolution verified by running firebase-tools 15.22.3's own resolver (`lib/hosting/config.js`) against the real `firebase.json` and `.firebaserc`, for every deploy this repo performs:

| command | resolves to |
|---|---|
| `--only hosting:walletrc --project komodo-wallet-official` | site `walletrc` |
| `--only hosting:walletrc --project komodo-wallet-preview` | site `komodo-wallet-preview` |
| `--only hosting:walletrc --project gleec-wallet-official` | site `gleec-wallet-official` |

Operator mistakes fail loudly rather than deploying somewhere unintended: a wrong `--only` aborts with *"Hosting site or target … not detected in firebase.json"*, a missing mapping with *"Deploy target … not configured for project …"*. The cost is that `walletrc` now names production too, so `--project` is what selects the site — documented in `docs/WEB_HOSTING_TOPOLOGY.md`, along with the fact that a bare `firebase deploy --project gleec-wallet-official` pushes whatever is in `build/web` straight to dex.gleec.com.

Two things the merge itself nearly got wrong, both now guarded:

- `git add -A` during a conflicted merge re-stages the `sdk` submodule from the stale working tree, silently reverting dev's newer pointer. Caught because the merged suite stopped compiling — dev's new tests reference SDK types that only exist in the newer submodule.
- Nothing asserted dev's coin-icon `Cache-Control` block, so restructuring this file could have dropped it without a test noticing. There is an assertion for it now.

Suite green with CI's dart-defines: **695 passed, 3 skipped, 0 failed**.

**Deploying production is a release, not a footnote.** It is on 0.9.4 from 2026-04-17; deploying current `dev` moves the whole web app forward by everything merged since, not just this fix. This PR makes production *possible* to deploy correctly and *checkable* afterwards; it does not make that call.

## Follow-ups (not in this PR)

- Confirm the real Banxa partner checkout host, then tighten the allowlist to exact hosts.
- Confirm the manual production deploy actually uses this repo's `firebase.json`. If it runs from a separate checkout, the new `gleec-wallet-official` entry is not the config in effect and the `headers` block has to be copied into whatever is. The verifier answers this after a deploy: wrapper checks passing while header checks fail means exactly that.
- `bitrefill_widget.js` interpolates URL parameters into the Bitrefill URL without encoding. It cannot change the origin (the base is hardcoded), so it is not the same class of issue, but it does let extra query parameters be injected into the provider URL.
- Pin the framed page's *subsequent* navigations, not just the first. Nothing re-validates the provider frame once `src` is assigned, so an open redirect on provider infrastructure could leave an attacker-controlled page rendered inside the buy dialog. It is origin-isolated, holds no delegated permissions, and its messages are dropped by the origin gate, so it is a phishing surface rather than an escalation — but a `frame-src https://*.banxa.com https://*.banxa-sandbox.com https://*.ramp.network` scoped to `/assets/assets/web_pages/**` would close it. Deferred for the same reason as the rest of the CSP: providers redirect inside the frame for 3DS and KYC, so it needs a staging soak before it can ship.
- The dead `main` → `prodrc` step in `firebase-hosting-merge.yml` should be removed or repointed; it cannot work as written. It is now dead in a second way: `prodrc` has no `.firebaserc` target mapping either, so it aborts at config resolution before reaching the API. That is pre-existing on `dev`, not introduced here.
- `.github/actions/firebase-hosting-deploy/action.yml` conflates target names with site IDs in two places — `hosting:channel:open --site "$TARGET"` and `details_url="https://$TARGET.web.app/"`. Both work today only because the target `walletrc` happens to be spelled like the RC site. Neither is on a path this PR changes (the live `details_url` is never consumed, and previews always run against `komodo-wallet-official`), but they are why the target was not renamed to something project-neutral while it was being generalised.


DeckerSU:
Why did the `macos/Runner/GoogleService-Info.plist` content change? Previously it had placeholders like THIS_IS_AUTOGENERATED, but now it contains real values from the komodo-wallet-preview project. Are you sure it's correct for any third-party builds to be compiled with these values?

Issue #3290, opened 2025-10-29
Document custom KDF commits builds


smk762:
Example command: 

```
dart run sdk/packages/komodo_wallet_cli/bin/update_api_config.dart   \
    --branch dev \
   --source mirror  \
   --config sdk/packages/komodo_defi_framework/app_build/build_config.json  \
   --output-dir sdk/packages/komodo_defi_framework/app_build/temp_downloads \
   -m 9aa41b4
```

We should include this somewhere in the docs folder and explain the flags (or link to the same in SDK repo)

AzazelSensei:
Opened https://github.com/GLEECBTC/gleec-wallet/pull/3522 — docs page for pinning a KDF commit with `update_api_config.dart -m`, plus the flag list.

PR #3522, opened 2026-09-01
Document custom KDF commit builds #3290

AzazelSensei:
I added a short docs page for pinning a KDF build to a specific commit.

It uses the `update_api_config.dart` command from #3290, lists the flags (`-m` / `--commit` is the one that pins), and links the SDK CLI README.

Fixes #3290


AzazelSensei:
Closing this after 15 days without a comment.

PR #3534, opened 2026-09-14
fix(security): remove private key sharing button

DeckerSU:
Removes the Share private keys action from the private-key export screen.

The unused share delivery path, its `share_plus` dependency in the export service, and share-specific are removed. Private keys can still be copied or, better, downloaded.


CharlVS:
Fixed a bug on the private key export on narrow screens.

